PR #88, opened 2014-08-07
Move pre-commit hook to pre-push hook

Dylnuge:
@azirbel @yayalice Let me know what you guys think. This was requested and I think it makes sense; we lose the verify on each commit but a lot of people have a workflow where they make dozens of intermediate commits and squash them all together at the end.

We should hold on merging until master is fixed.

---
Commit hooks prevent people who use intermediate commit workflows from
committing their temporary or incomplete changes. This in turn promotes
using --no-verify and similar options to circumvent the hooks entirely.
Checking test integrity on push loses the ability to verify each commit
but works better within the workflow of people using intermediate
commits.

Author: Dylan Nugent


Dylnuge:
@igillis I'm closing this because we decided against doing this so long as tests weren't even passing and we can revisit how to do commit hooks once this library actually has functional tests.

PR #139, opened 2014-11-06
Allow popover to start in expanded state

bjcohen:
Add option to `popover_link.coffee` so that a user of the link widget can specify that the popover should be shown immediately after the page renders.

@igillis @azirbel


cyril-sf:
@bjcohen can we delete that branch?

PR #140, opened 2014-11-10
Select - reset and native html <select> behavior

shturm:
Latest UX/Usability test showed users are confused with search behavior.
- Reset Select query when dropdown hides
- Add optional querying only from the beggining of the label

They expect search criteria to be reset when they open dropdown to change selection, because otherwise they manually have to clear it with `backspace` key. The test included picking automobile brands. When entering one letter users expected to see brands starting with that letter, similar to the native `<select>` behavior. E.g. "P" should have shown "PEUGEOT", but it showed "ALPINA" and other brands 1st because brands were initially sorted alphabetically. This behavior is optionable and turned off by default.


igillis:
@shturm did you ever get a chance to make those changes?

shturm:
Hi, unfortunately not, but will do by the end of the month. I'm quite busy lately.

mixonic:
Closing this due to inactivity 👍

PR #156, opened 2014-12-28
Memory leak - destroy content proxy

cyril-sf:
Make sure to destroy `contentProxy` so that observers get torn down.

Also optimize the way this property gets created.

@addepar-andy  @igillis  @Dylnuge @azirbel


src/select.coffee
```diff
@@ -107,6 +107,16 @@ Ember.AddeparMixins.ResizeHandlerMixin,
...
   selectMenuView: null
 
+  # This doesn't clean correctly if `optionLabelPath` changes
+  willDestroy: ->
+    propertyName = 'contentProxy'
+    if @cacheFor propertyName
+      contentProxy = @get propertyName
+      contentProxy.destroy()
+
+    @_super.apply this, arguments
```
ebryn: might as well be consistent with `@_super()`

ebryn:
LGTM otherwise

PR #161, opened 2015-01-08
Add a loading message to the select widget dropdown

ebryn:
A good first pass, might want to make this more customizable.


igillis:
I think this is fine for now. If someone wants to add a custom view there it should be an easy PR. 

LGTM

PR #162, opened 2015-01-10
Modals don't crash when activating Ember.Widgets.DISABLE_ANIMATIONS

cyril-sf:
@yayalice @addepar-andy @igillis


addepar-andy:
lgtm

cyril-sf:
There are actually more problems I need to take care of but I need to focus on something else.

I'll close that PR for the moment.

PR #165, opened 2015-01-14
Tab through fields inside modal only

thangdinh:
I break my ongoing project into smaller branches so that it's easier to track and review. Here is the branch for tab inside modal only. When a modal is popup, pressing TAB only focus around the fields inside the modal. 

Please help me review it @cyril-sf @Dylnuge . Thanks.


src/mixins.coffee
```diff
@@ -64,3 +64,75 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
   _removeDocumentHandlers: ->
     $(@get('bodyElementSelector')).off "click", @_clickHandler
     @_clickHandler = null
+
+Ember.Widgets.TabbableModal = Ember.Mixin.create
+  currentFocus: null
+
+  _focusTabbable: ->
+     # Set focus to the first match:
```
azirbel: This comment makes sense to me, but I find the logic below very confusing. What if you made a function `setFocusToElement(domElement)`, then build an `if...else` block to mirror your comment. So for example: `if @get 'currentFocus'; setFocusToElement(...); else if @$('[autofocus']) is not null; setFocusToElement(...)`. Does that work?

src/mixins.coffee
```diff
@@ -64,3 +64,75 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+     # 1. The saved focused element
+     # 2. First element inside the dialog matching [autofocus]
+     # 3. Tabbable element inside the content element
+     # 4. The close button (has class "close")
+    hasFocus = [@get 'currentFocus']
+    if hasFocus[0] == null or not $.contains(@$()[0], hasFocus[0])
+      hasFocus = @$( "[autofocus]" )
```
azirbel: Should use single quotes consistently

src/mixins.coffee
```diff
@@ -64,3 +64,75 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+  currentFocus: null
+
+  _focusTabbable: ->
+     # Set focus to the first match:
+     # 1. The saved focused element
+     # 2. First element inside the dialog matching [autofocus]
+     # 3. Tabbable element inside the content element
+     # 4. The close button (has class "close")
+    hasFocus = [@get 'currentFocus']
+    if hasFocus[0] == null or not $.contains(@$()[0], hasFocus[0])
```
azirbel: Should use `is` instead of `==`

src/mixins.coffee
```diff
@@ -64,3 +64,75 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+      focusElement = hasFocus[0]
+      if focusElement.className.indexOf("close") > -1
+        # if we have more than two tabbable objects, we do not want to tab
+        # to close button
+        # while if we do not have any choice, the close button is chosen
+        if hasFocus.length > 1
+          focusElement = hasFocus[1]
+      focusElement.focus()
+      @set 'currentFocus', focusElement
+
```
azirbel: extra newline

src/mixins.coffee
azirbel: This applies to a lot of places in this PR, make sure to check

src/mixins.coffee
```diff
@@ -64,3 +64,75 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+        if (event.target == last and not event.shiftKey)
+          first.focus()
+          @set 'currentFocus', first
+          event.preventDefault()
+        else if (event.target == first and event.shiftKey)
+          @set 'currentFocus', last
+          last.focus()
+          event.preventDefault()
+        else
+          @set 'currentFocus', tabbableObjects[_index+1]
```
azirbel: does this respect SHIFT+TAB?

src/mixins.coffee
```diff
@@ -64,3 +64,75 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+  # inside the modal. Remove the close button from the loop
+  keyDown: (event) ->
+    return if event.isDefaultPrevented()
+
+    if event.keyCode == @KEY_CODES.ESCAPE and @get 'escToCancel'
+      @doCancelation()
+      event.preventDefault()
+    else if event.keyCode == @KEY_CODES.TAB
+      # tabbable objects list without close button
+      tabbableObjects = @$(":tabbable").not('.close')
```
azirbel: So you can't tab to the close button? Why?

src/mixins.coffee
```diff
@@ -64,3 +64,75 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+    if not @_checkContainingElement(@$(':focusable'), event.target)
+      @_focusTabbable()
+
+  # capture the TAB key and make a cycle tab loop among the tabbable elements
+  # inside the modal. Remove the close button from the loop
+  keyDown: (event) ->
+    return if event.isDefaultPrevented()
+
+    if event.keyCode == @KEY_CODES.ESCAPE and @get 'escToCancel'
+      @doCancelation()
```
azirbel: this is very messy. Is it really not possible to use `@send` or `@sendAction`? If not - the real problem is our confirm/cancel architecture in the modal component. I think the recommended way is to put all the logic in the `doCancelation()` function and have the action call that.

This isn't all that bad, though.

azirbel: At the very least you should put a comment at the top of this mixin indicating the functions that must be defined by any controller using this mixin

src/mixins.coffee
azirbel: Tried it out in the ember-widgets demo app and it looks like you can indeed tab to the close button. I don't understand how.

src/mixins.coffee
```diff
@@ -64,3 +64,75 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+        return yes
+    return no
+
+  click: (event) ->
+    isActive = $.contains(@$()[0], event.target)
+    if not isActive
+      @hide() unless @get('enforceModality')
+      @_focusTabbable()
+    if not @_checkContainingElement(@$(':focusable'), event.target)
+      @_focusTabbable()
```
azirbel: This means that if you click inside the modal, on something that's not tabbable (like the modal itself), the modal will focus the first tabbable element again. I'm not sure that's the best behavior - typically when I click, I expect that the currently focused element will lose focus. You can always get focus back again by pressing TAB.

src/mixins.coffee
```diff
@@ -64,3 +64,75 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+    if event.keyCode == @KEY_CODES.ESCAPE and @get 'escToCancel'
+      @doCancelation()
+      event.preventDefault()
+    else if event.keyCode == @KEY_CODES.TAB
+      # tabbable objects list without close button
+      tabbableObjects = @$(":tabbable").not('.close')
+
+      _currentFocus = $(document.activeElement)?[0]
+
+      _index =tabbableObjects.index _currentFocus
```
azirbel: space after =

src/mixins.coffee
```diff
@@ -64,3 +64,75 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+      if _index == -1
+        @_focusTabbable()
+      if (tabbableObjects.length > 0)
```
azirbel: no need for parens

src/mixins.coffee
azirbel: also, can you explain how this actually works? I don't see you calling `.focus()` anywhere.

azirbel:
It's frustraing that there isn't a good way for me to try this out and confirm that it works / see if the behavior feels right. I think we should either (1) write tests for this (if I recall correctly, ember-widgets has a test framework; tests just fail at the moment?) or (2) add another modal to the modal examples page in the demo app - this modal would be more complex and would have a few tabbable fields. Then we can at least verify that this feature works when other people submit patches/changes in the future.

azirbel:
Done reviewing for now; will check back when you respond to comments! It looks good and I'm excited we'll be able to tab through modals soon. Thanks for writing this =)

thangdinh:
Thanks for a very detailed review @azirbel :+1:  I am on the UI2 Dash now. I'll respond to you one by one when I finish these. Thanks :)

src/mixins.coffee
```diff
@@ -64,3 +64,75 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+     # 1. The saved focused element
+     # 2. First element inside the dialog matching [autofocus]
+     # 3. Tabbable element inside the content element
+     # 4. The close button (has class "close")
+    hasFocus = [@get 'currentFocus']
+    if hasFocus[0] == null or not $.contains(@$()[0], hasFocus[0])
+      hasFocus = @$( "[autofocus]" )
+    if hasFocus.length == 0
+      hasFocus = @$( ":tabbable" )
+    if hasFocus.length > 0
```
cyril-sf: You can do this in one line

`if focusElement = hasFocus[0]`

src/mixins.coffee
cyril-sf: Also no need for `> 0` as anything above 0 will be truthy

src/modal.coffee
```diff
@@ -96,6 +99,10 @@ Ember.Component.extend Ember.Widgets.StyleBindingsMixin,
 
   didInsertElement: ->
     @_super()
+    # Make sure that after the modal is rendered, set focus to the first
+    # tabbable element
+    Ember.run.schedule 'afterRender', this, ->
```
cyril-sf: You will need to cancel it in case someone destroys the modal right after inserting it in the DOM

src/mixins.coffee
thangdinh: Yes it does. But I cannot make the code shorter or clearer with that function. Because we still have to check through different arrays of focusable elements and come down to that road to replace each of 2-3 lines of code with a function call. I am not sure if I understand it right.

src/mixins.coffee
thangdinh: Is it from style guide or is there any other reason making is better than ==? I was also curious about that but hasn't got good answer.

src/mixins.coffee
thangdinh: Is it easy to get confused with `if focusElement == hasFocus [0]`? I usually try to avoid putting assignment in condition statements (the valid ones) so that i will not make mistakes with the invalid ones.

src/mixins.coffee
thangdinh: It is a bit different: 
It will focus on the latest focused item (if there is one, otherwise, the first tabbable one, as you described) if we click on an element not tabbable.
But I get your point. I think it's a good idea. I will check with some people to see which one they like.

src/mixins.coffee
thangdinh: Got it :+1:

src/mixins.coffee
thangdinh: When I implemented this one, I double check with some websites having popup, it does seem common. I do not know the underlying reasons but I guess if it is meant for cancelation, there should be cancel button (or ESC). The close button normally is relatively small and it is designed for mouse click? I then double checked with jQuery and I saw them implementing it like that. So I followed them.

thangdinh: Btw, i just checked mine. It doesn't tab through close button... hmm...

src/mixins.coffee
thangdinh: Yes it does respect Shift+Tab. The concept is: we do not change the behavior of tabbing, it is browser built-in. We only change the behavior is when users hit Tab while they are on the last element, or they hit Shift-Tab on the first element of the modal. The only thing we need to do is keep track of what is the current focused item. So that we can guide it back when necessary.

src/modal.coffee
thangdinh: :+1:

thangdinh:
Does anyone having idea to make a more complex and "meaningful" modal to test this? @azirbel @cyril-sf @igillis

azirbel:
@thangdinh here are some options. (1) add a complex modal to the demo app; incorporate other widgets like the select widget etc. (2) test against another app using `bower link` and make sure the behavior is what you'd expect. If possible, do both.

cyril-sf:
We desperately need to add tests :/

thangdinh:
@cyril-sf It seems like I cannot trigger the tab key to build tests for this. Do you have any suggestion?

igillis:
@thangdinh You can trigger the tab key, but the issue is that triggering it with js doesn't actually do the browser's default actions, correct? So you have to just manually focus the element.

thangdinh:
@igillis if so the test will look kind of similar to the code implemented in the branch. There will be not much value added by the test, is it correct?

igillis:
Wait it looks like your code already captures the tab keypress and overrides default behaviors so you should be able to test it?

thangdinh:
@igillis no, my code is only to check when the focus is at the ending element of the loop to return the focus back to the 1st element. The tab behavior is preserved as default.

igillis:
@thangdinh can you push your tests so I can check them out and see what you're doing?

src/mixins.coffee
```diff
@@ -64,3 +77,81 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+  currentFocus: null
+
+  _focusTabbable: ->
+     # Set focus to the first match:
+     # 1. The saved focused element
+     # 2. First element inside the dialog matching [autofocus]
+     # 3. Tabbable element inside the content element
+     # 4. The close button (has class "close")
+    hasFocus = [@get 'currentFocus']
+    if hasFocus[0] is null or not $.contains(@$()[0], hasFocus[0])
```
cyril-sf: `if not @$().has( hasFocus[0] ).length` should be enough

src/mixins.coffee
```diff
@@ -64,3 +77,81 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+      @set 'currentFocus', focusElement
+
+  _checkContainingElement: (containers, element) ->
+    for containerItem in containers
+      if containerItem is element or $.contains(containerItem, element)
+        return yes
+    return no
+
+  click: (event) ->
+    if @$()[0].contains(event.target) or not document.contains(event.target)
```
cyril-sf:
```cs
@$().has( event.target )
```

?

cyril-sf: How can the document not contain the target?

src/mixins.coffee
```diff
@@ -64,3 +77,81 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+    return if event.isDefaultPrevented()
+
+    if event.keyCode == @KEY_CODES.ESCAPE and @get 'escToCancel'
+      @send 'sendCancel'
+      event.preventDefault()
+    else if event.keyCode == @KEY_CODES.TAB
+      # tabbable objects list without close button
+      tabbableObjects = @$(":tabbable").not('.close')
+
+      _currentFocus = $(document.activeElement)?[0]
```
cyril-sf: `_currentFocus = document.activeElement` should be enough

src/mixins.coffee
```diff
@@ -64,3 +77,81 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+
+      _index = tabbableObjects.index _currentFocus
+      if _index == -1
```
cyril-sf: `is`

src/mixins.coffee
thangdinh: @cyril-sf Because the target is removed from the document after the click event is triggered. There are several places where we replace the current element by another element for styling purpose.

thangdinh:
@igillis I just pushed the test. Would you please help me have a look at it? Right now, I am using jQuery event to trigger shift+TAB. I will change to triggerEvent after @cyril-sf upgrades Ember version

src/mixins.coffee
```diff
@@ -64,3 +77,95 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+        if hasFocus.length > 1
+          focusElement = hasFocus[1]
+      focusElement.focus()
+      @set 'currentFocus', focusElement
+
+  _checkContainingElement: (containers, element) ->
+    for containerItem in containers
+      if containerItem is element or $.contains(containerItem, element)
+        return yes
+    return no
```
igillis: nit: don't need the return

src/mixins.coffee
```diff
@@ -64,3 +77,95 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
...
+
+  click: (event) ->
+    @_super()
+    if event.target is event.currentTarget
```
igillis: nit: can you add a comment here

igillis:
lgtmbc

src/mixins.coffee
thangdinh: @igillis did you mean the default returned value will be `no`? Should we make it more explicit for easier reading?

igillis: I meant you can just do `no` instead of `return no`

thangdinh: @igillis Is it ok if I prefer having it explicit?

igillis: let's just be consistent with the rest of the codebase(s)

thangdinh:
Merged

PR #166, opened 2015-01-15
Improve usability of Select component

thangdinh:
Basically, I would like to make select component now tabbable.  I also changed the behavior of pressing any key will show the dropdown, because it seems to be uncommon to show the dropdown when we press special keys such as: shift, capslock, F1,F2...
Please help me review it @igillis @azirbel @hanakimis 
Thank you.


src/css/select.less
```diff
@@ -23,6 +24,13 @@
   position: relative;
   display: inline-block;
   outline: none;
+
+  &:focus {
+    .ember-select-choice {
```
hanakimis: can we put these on the same line, so it reads: "&:focus .ember-select-choice"? It makes it easier to read.

Thanks!

hanakimis:
CSS LGTM, besides comment : )

src/select.coffee
```diff
@@ -341,7 +342,15 @@ Ember.AddeparMixins.ResizeHandlerMixin,
 
   keyDown: (event) ->
     # show dropdown if dropdown is not already showing
-    return @set('showDropdown', yes) unless @get 'showDropdown'
+    # and the keycode should be one of special keys or common characters
+    # 'A'..'Z', 'a..z','0..9'
+    if event.keyCode in [@KEY_CODES.ENTER, @KEY_CODES.SPACEBAR] or
+    event.keyCode in [@KEY_CODES.DOWN, @KEY_CODES.UP] or
```
azirbel: indentation is weird here; these should be indented twice, right?

src/select.coffee
```diff
@@ -349,7 +358,14 @@ Ember.AddeparMixins.ResizeHandlerMixin,
...
 
   escapePressed: (event) ->
-    @send 'hideDropdown'
+    if @get('showDropdown') is yes
+      @send 'hideDropdown'
+      @$().focus()
+      event.preventDefault()
+
+  tabPressed: (event) ->
+    if @get('showDropdown') is yes
```
azirbel: `if @get('showDropdown')`

azirbel: Or, more concisely: `@send 'hideDropdown' if @get('showDropdown')`

azirbel:
Thanks for this @thangdinh 

Tested it out and I think it's a clear improvement over the current behavior, but a few scenarios are still a bit weird:

* In my opinion, hitting ESC should unfocus the component
* In the multi-select component, pressing ENTER seems to remove focus from the component, so you can't type more
* Actually, the multi-select component seems pretty broken, I've gotten it into a state where it doesn't filter when I type ... looks like this happens when I select one option, then click back into the box to type another. This might not be your fault.

I don't understand what you mean by "it seems to be uncommon to show the dropdown when we press special keys such as: shift, capslock..." For me it looks like the dropdown is always shown when the component comes in focus, even before characters are typed. Why do we need the special cases for common characters?

src/select.coffee
bjcohen: @azirbel Do we have a style recommendation for something like this? Indenting subsequent lines of the condition means that the body of the conditional needs to be indented to 4 spaces, which is also weird.

addepar-andy: for this specific case, i'd make a static map with all the keys that match, and then check that map.

for general cases, when you have 40 different clauses in your `if` statement, its better to put each clause into a variable name that makes sense and then use those names, so that the statement is more readable.

thangdinh:
@azirbel Thanks Alex for your detailed comments. Here are my thoughts:
+ ESC (on select) should close the dropdown, shouldn't unfocus the component, on forms there should be at least one component gets focused. On popular websites such as gmail, ebay... it appears that way.
+ I tested it on multi select before I left on the other branch. But i haven't tested it this time. It's my bad. I'll do it.
About the dropdown's behavior when the select gets focused, similar to my first point, on popular websites the dropdown only appears when you hit some special keys such as ENTER, SPACEBAR, arrow keys or hit the first letter of one option inside the list. Apparently, I think it's a good behavior. We shouldn't show dropdown when we focus on it because it looks interrupting (UI/UX). Imaging that when you already selected one option, and you navigate back and forth (tab, shift tab) it keeps popping out for no good reason. Anyway, UI/UX is a very debatable topic so I guess our product/designing team will decide on how it behaves.
I don't have much experience with multi-select component, so I'm not sure if it follows the same behavior as above or not. Do you have any websites in mind having such components?

azirbel:
* Ok, I'm ok with that ESC behavior. But when I run your branch and use the select or multi-select, I seem to lose focus and the dropdown stays open - are you sure it's working the way you described?
* The second behavior also makes sense to me. So you're saying if I TAB to a select component, it should focus but not open the dropdown. Then when a key is pressed, it should open the dropdown. I think this behavior sounds good, but again when running your branch I'm seeing the dropdown open as soon as I TAB to the select component, so it doesn't seem to be working for me.

thangdinh:
@azirbel:  I'm checking on this. And fixing the multi-select component problem. I tested again the cases you mentioned, and it runs fine on select component. I am not sure why it didn't work on your computer.

thangdinh:
Could anyone help me review this PR again? Thanks.
@hanakimis, @azirbel, @igillis, @Dylnuge

dist/css/ember-widgets.css
```diff
@@ -45,6 +45,14 @@
   display: inline-block;
   outline: none;
 }
+.ember-select:focus .ember-select-choice,
```
hanakimis: Are you sure you're supposed to edit the css files in dist, rather than src?

igillis: this is the compiled version of the changes in `select.less`, and he does need to check this in

src/css/select.less
```diff
@@ -6,6 +6,8 @@
 -------------------------------------------------- */
 // Note: Important that this is synchronized with the list-view rowHeight
 @rowHeight: 26px;
+@input-border-focus: #4d90fe;
```
azirbel: can we use variables here? If not, comment on where the values came from?

azirbel: wow, I completely misread this. deleting my comment.

src/multi_select.coffee
```diff
@@ -47,10 +47,27 @@ Ember.Widgets.MultiSelectComponent = Ember.Widgets.SelectComponent.extend
     @get('placeholder') or @get('persistentPlaceholder')
   .property 'query', 'placeholder', 'persistentPlaceholder', 'selections.length'
 
+  # add css class to multiselect to highlight border
```
azirbel: no need for this comment or the one on line 54, since your function names are good.

src/multi_select.coffee
```diff
@@ -47,10 +47,27 @@ Ember.Widgets.MultiSelectComponent = Ember.Widgets.SelectComponent.extend
...
+    # we want to focus on search input when dropdown is opened. We need to put
+    # this in a run loop to wait for the event that triggers the showDropdown
+    # to finishes before trying to focus the input. Otherwise, focus when be
+    # "stolen" from us.
+    showDropdownDidChange: Ember.observer ->
+      # when closing, don't need to focus the now-hidden search box
+      if @get('parentView.showDropdown')
+        Ember.run.schedule 'afterRender', this, ->
+          @$().focus() if (@get('_state') or @get('state')) is 'inDOM'
+    , 'parentView.showDropdown'
```
azirbel: add newline between this and `placeholder`

src/multi_select.coffee
```diff
@@ -98,12 +115,30 @@ Ember.Widgets.MultiSelectComponent = Ember.Widgets.SelectComponent.extend
     @set 'values', Ember.A [] unless @get('values')
 
   deletePressed: (event) ->
-    if event.target.selectionStart == 0
+    # check if the cursor is at the beginning and no text is selected
+    if event.target.selectionStart == 0 and event.target.selectionEnd == 0
```
azirbel: not your fault since this code existed before, but use `is` instead of `==`

src/multi_select.coffee
azirbel: that said, `addBorderToMultiSelectContainer` and `removeBorderFromMultiSelectContainer` would be better

src/multi_select.coffee
```diff
@@ -98,12 +115,30 @@ Ember.Widgets.MultiSelectComponent = Ember.Widgets.SelectComponent.extend
...
+  focusOut: (event) ->
+    @removeBorderMultiSelectContainer()
+
+  keyDown: (event) ->
+    @_super(event)
+    activeElement = $(document.activeElement)[0]
+
+    # dynamically add css style to highlight the border if any elements inside
+    # the multiselect is focused
+    if @$()[0].contains(activeElement) or @$()[0]==activeElement
```
azirbel: use `is`

src/select.coffee
```diff
@@ -107,6 +107,21 @@ Ember.AddeparMixins.ResizeHandlerMixin,
   # possibly says 'create item' or something along that line
   selectMenuView: null
 
+  # a map of accepted keys to show dropdown when being pressed
+  acceptedKeys: null
+
+  # set up a map for accepted keys from multiple ranges of them
+  # these are keys to show dropdown when being pressed
+  createAcceptedKeySet: ->
```
azirbel: why not make this a computed property? `acceptedKeys: Ember.computed -> ... .property()`

src/select.coffee
```diff
@@ -340,8 +360,13 @@ Ember.AddeparMixins.ResizeHandlerMixin,
...
 
   keyDown: (event) ->
-    # show dropdown if dropdown is not already showing
-    return @set('showDropdown', yes) unless @get 'showDropdown'
+    # show dropdown if it is not already showing
+    # and the keycode should be in the list of accepted keys to show dropdown
+    # [Spacebar, Enter, Up, Down, 'A'..'Z','a..z','0..9']
+    acceptedKeys = @get 'acceptedKeys'
+    if acceptedKeys[event.keyCode]
+      return @set('showDropdown', yes) if not @get 'showDropdown'
```
azirbel: nitpick, but `if not` -> `unless`

src/select.coffee
```diff
@@ -64,10 +64,10 @@ Ember.AddeparMixins.ResizeHandlerMixin,
   placeholder:        undefined
   disabled: no
 
-  # we need to set tabindex so that div responds to key events
   highlightedIndex: -1
 
-  tabindex: -1
+  # we need to set tabindex so that div responds to key events
+  tabindex: 0
```
azirbel: this change is good, but you should add `tabindex: -1` to the MultiSelectComponent, because right now you have to press tab twice (once for the component and once for the text field inside). I just tested with `tabindex: -1` and it seems to work as expected.

azirbel:
Not sure why, but "hitting esc to close the dropdown and re-select the component" is not working for me. When I hit esc, I lose focus and the dropdown stays open

src/multi_select.coffee
azirbel: is this code necessary? I tried deleting it and the border still seemed to appear/disappear as expected. If it's necessary please add a comment explaining the case that breaks without it

azirbel:
LGTMBC, a few comments are important so let me know if you have questions

azirbel:
Actually just noticed something else, the cursor is a bit weird once you select one element in multi-select. So you select one thing and focus goes back to the component, at which point you CAN type, but it doesn't *look* like you can type because there is no cursor (focus is on the component, not the text field).

thangdinh:
Thanks @azirbel. I addressed all of the comments you mentioned. However, I could not reproduce this issue: "When I hit esc, I lose focus and the dropdown stays open". When you have time, would you please help me double check it? Steps to reproduce it will be appreciated. Thanks.

azirbel:
Sure. (1) Click inside multi select component; (2) type "af". Dropdown should appear. (3) hit ESC. focus is lost and the dropdown stays open.

Just got the above with the latest code from your branch.

On the other hand, the current state is better than what's on master (we could consider this an existing bug). If the fix is out of scope or too hard, you can skip it.

By the way, the tab behavior is really awesome. Thanks for working on it.

thangdinh:
Thanks @azirbel. I still cannot reproduce the ESC behavior. I don't know why. 
Could you help me double check this behavior when you have time, @igillis? Thanks.

igillis:
I just checked out this branch and saw the following in both Chrome 41.0.2272.76 (64-bit) and FF 35.0.1 on Mac OS X 10.9.5.

1) Click inside the multi select
2) Type 'af'. Dropdown appears.
3) Hit ESC.
4) Dropdown closes, focus remains on the button, 'af' still present in the textbox

Similarly for the normal select, except when you re-open the dropdown 'af' isn't present. @azirbel I wonder if you have a different version of something? Would be curious where it's failing for you -- do you reach `escapePressed`?

igillis:
@azirbel: I get your behavior unless I disable my vimium plugin...any chance you're running that?

azirbel:
........you are a debugging god

thangdinh:
Thanks @igillis. You are awesome!

thangdinh:
I just added keyboard interaction tests. Would you please help me review it? Thanks.
@igillis @cyril-sf @azirbel

igillis:
@thangdinh and I spoke offline -- can you comment again when you get rid of the unnecessary `wait`s

thangdinh:
@igillis done

igillis:
I'm getting
```
FAILED  - Test keyboard interaction	Died on test #2     at /Users/ian.gillis/ember-widgets2/tests/unit/select.js:59:1: undefined is not a function
	TypeError: undefined is not a function
	    at /Users/ian.gillis/ember-widgets2/tests/unit/select.js:70:36
```
on my machine -- does it work on yours?

thangdinh:
My bad. Could you please try it again?

src/multi_select.coffee
```diff
@@ -98,12 +107,26 @@ Ember.Widgets.MultiSelectComponent = Ember.Widgets.SelectComponent.extend
...
+  focusIn: (event) ->
+    @addBorderMultiSelectContainer()
+
+  focusOut: (event) ->
+    @removeBorderMultiSelectContainer()
+
+  keyDown: (event) ->
+    @_super(event)
+    activeElement = $(document.activeElement)[0]
+    if @$()[0].contains(activeElement) or @$()[0]==activeElement
```
igillis: nit: use `is` instead of == and use proper spacing

nit: local variable with meaningful name for `@$()[0]`

src/select.coffee
```diff
@@ -195,6 +214,9 @@ Ember.AddeparMixins.ResizeHandlerMixin,
       if @get('parentView.showDropdown')
         Ember.run.schedule 'afterRender', this, ->
           @$().focus() if (@get('_state') or @get('state')) is 'inDOM'
+      # clear the query string when dropdown is hidden
```
igillis: :+1:

igillis:
Nice usability win @thangdinh. Thanks for your patience and for adding some much-needed tests.

LGTM besides comments (you can merge after fixing my comments)

src/multi_select.coffee
```diff
@@ -98,12 +107,27 @@ Ember.Widgets.MultiSelectComponent = Ember.Widgets.SelectComponent.extend
...
   deletePressed: (event) ->
-    if event.target.selectionStart == 0
+    if event.target.selectionStart is 0 and event.target.selectionEnd is 0
       @removeSelectItem(@get('selections.lastObject'))
 
   removeSelectItem: (item) ->
     @get('selections').removeObject item
 
+  focusIn: (event) ->
+    @addBorderMultiSelectContainer()
```
cyril-sf: I would probably set a boolean ( `hasFocus` ) to `yes` / `no` and use `bind-attr` in the template to toggle the class

cyril-sf:
```hbs
<div {{bind-attr class="hasFocus:ember-select-multi-focus"}}>
```

src/select.coffee
```diff
@@ -195,6 +214,9 @@ Ember.AddeparMixins.ResizeHandlerMixin,
...
+      else
+        @set 'parentView.query', ''
```
cyril-sf: it seems like you can access the query through the `value` property

src/select.coffee
```diff
@@ -340,13 +367,18 @@ Ember.AddeparMixins.ResizeHandlerMixin,
...
 
   keyDown: (event) ->
-    # show dropdown if dropdown is not already showing
-    return @set('showDropdown', yes) unless @get 'showDropdown'
+    # show dropdown if it is not already showing
+    # and the keycode should be in the list of accepted keys to show dropdown
+    # [Spacebar, Enter, Up, Down, 'A'..'Z','a..z','0..9']
+    acceptedKeys = @get 'acceptedKeys'
+    if acceptedKeys[event.keyCode]
+      return @set('showDropdown', yes) unless @get 'showDropdown'
```
cyril-sf: why not test both conditions in the `if` statement?

src/select.coffee
```diff
@@ -354,14 +386,23 @@ Ember.AddeparMixins.ResizeHandlerMixin,
   deletePressed: Ember.K
 
   escapePressed: (event) ->
-    @send 'hideDropdown'
+    if @get('showDropdown')
+      @send 'hideDropdown'
+      @$()?.focus()
```
cyril-sf: how can `@()` be null?

tests/unit/multi_select.coffee
```diff
@@ -0,0 +1,67 @@
...
+    highlightedComponent = find '.ember-select-multi', multiSelectComponent
+
+    multiSelectComponent.focus()
+    wait().then ->
+      # test pressing ENTER key to open dropdown
+      keyEvent(multiSelectComponent, 'keydown', 13)
+    .then ->
+      ok isVisible(find '.ember-select-results', multiSelectComponent, 'Dropdown list should appear')
+
+    .then ->
```
cyril-sf: it should not be needed

tests/unit/multi_select.coffee
```diff
@@ -0,0 +1,67 @@
...
+      keyEvent(multiSelectComponent, 'keydown', 13)
+
+    .then ->
+      ok isHidden(find '.ember-select-results', multiSelectComponent, 'Dropdown list should be hidden')
+
+      # test if selected Item is actually selected
+      resultItems = find '.ember-select-search-choice', multiSelectComponent
+      currentText = $(resultItems[resultItems.length - 1]).text().trim()
+      # trim the special character "x" at the end of the select choice text
+      equal(currentText.substring(0, currentText.length - 2), selectedText)
```
cyril-sf: Adding a description to the assertion helps to understand the test without reading the actual code

tests/unit/multi_select.coffee
```diff
@@ -0,0 +1,67 @@
...
+
+      # test if dropdown appears when we start typing letter ('a' is input here)
+      keyEvent(multiSelectComponent, 'keydown', 97)
+
+    .then ->
+      ok isVisible(find '.ember-select-results', multiSelectComponent, 'Dropdown list should appear')
+
+      keyEvent(multiSelectComponent, 'keydown', 27)
```
cyril-sf: which key is 27?

cyril-sf:
This is looking good. I'm happy to see some tests added :smile: 

Let me know if you have some questions about my comments.

src/multi_select.coffee
thangdinh: :+1: Thank you.

src/select.coffee
thangdinh: Do you mean doing this instead? 
```
@set 'value', ''
```

src/select.coffee
thangdinh: It is somehow null in the scenario where we put it on a carousel popover, and after hitting Enter, the carousel select disappears, and @$() is null. So I just wanted to make sure it not happening for ESC too

cyril-sf: I see.

It's probably cleaner to check the state of the component ( `if @get('isDestroyed') or @get('isDestroying')` ) at the beginning of the event handler

thangdinh:
could anyone help me review one more round? @cyril-sf @azirbel @igillis 
Thanks.

src/mixins.coffee
```diff
@@ -1,3 +1,16 @@
+Ember.Widgets.KeyboardHelper = Ember.Mixin.create
+  KEY_CODES:
+    BACKSPACE: 8,
```
azirbel: you could delete commas here since it's CS. nitpick.

src/multi_select.coffee
```diff
@@ -101,12 +114,33 @@ Ember.Widgets.MultiSelectComponent = Ember.Widgets.SelectComponent.extend
...
 
   deletePressed: (event) ->
-    if event.target.selectionStart == 0
+    if event.target.selectionStart is 0 and event.target.selectionEnd is 0
       @removeSelectItem(@get('selections.lastObject'))
+      event.preventDefault()
 
   removeSelectItem: (item) ->
+    # set the focus back to the searchView because this item will be removed
+    currentDropDownState = @get('showDropdown')
```
azirbel: another nitpick, but this is a weird name. Something like `dropdownIsShowing` would make line 125 read better.

src/multi_select.coffee
```diff
@@ -101,12 +114,33 @@ Ember.Widgets.MultiSelectComponent = Ember.Widgets.SelectComponent.extend
...
+  escapePressed: (event) ->
+    if @get('showDropdown')
+      @focusTextField()
+      @send 'hideDropdown'
+      event.preventDefault()
+
+  enterPressed: (event) ->
+    @_super(event)
+    @focusTextField()
+    @set('showDropdown', yes) unless @get('showDropdown')
```
azirbel: if it's already `yes`, there should be no harm setting it to `yes`, right? Can we remove the `unless`?

azirbel: same below in `click`. I wonder if I'm missing something?

src/select.coffee
```diff
@@ -44,7 +44,12 @@ Ember.Widgets.SelectOptionView = Ember.ListItemView.extend
     # propagate the click event
     # if the dropdown is expanded and we select something, don't propagate
     if @get('controller.showDropdown')
-      @get('controller').send 'hideDropdown'
+      # Keep showing dropdown if it is multi select component
+      unless @get('controller.layoutName') is 'multi-select'
```
azirbel: this is pretty bad, we should definitely find a way to override `click` in `multi-select` so that `select` doesn't have to have logic for `multi-select`.

I'd recommend refactoring lines 40-42 into a `selectContent` function, calling that in the `click` handler for both `select` and `multi-select`, and then having each component handle its own logic for hiding the dropdown. You could then remove the `@_super` call from the `multi-select` click handler.

src/select.coffee
```diff
@@ -346,11 +371,26 @@ Ember.AddeparMixins.ResizeHandlerMixin,
...
-    # show dropdown if dropdown is not already showing
-    return @set('showDropdown', yes) unless @get 'showDropdown'
+    return if @get('isDestroyed') or @get('isDestroying')
+    @setFocus()
+    # show dropdown if it is not already showing
+    # and the keycode should be in the list of accepted keys to show dropdown
+    # [Spacebar, Enter, Up, Down, 'A'..'Z','a..z','0..9']
+    acceptedKeys = @get 'acceptedKeys'
+    if acceptedKeys[event.keyCode] and not @get 'showDropdown'
+      return @set('showDropdown', yes)
```
azirbel: the result of a `@set` isn't very interesting; I think you should `set` and then `return` on a new line.

src/select.coffee
azirbel: you can do this in many places in the code.

tests/helpers/test_helpers.coffee
```diff
@@ -117,6 +121,7 @@ Ember.Test.registerHelper 'mouseDown', _mouseDown
 Ember.Test.registerHelper 'isPresent', _isPresent
 Ember.Test.registerHelper 'isNotPresent', _isNotPresent
 Ember.Test.registerHelper 'isVisible', _isVisible
+Ember.Test.registerHelper 'isFocus', _isFocus
```
azirbel: should rename to `isFocused`

tests/unit/multi_select_test.coffee
```diff
@@ -0,0 +1,96 @@
...
+  multiSelect.set 'selections', ['bar', 'baz']
+  multiSelect.set 'query', 'ba'
+  equal(multiSelect.get('preparedContent').length, 1)
+  equal(multiSelect.get('preparedContent')[0], 'barca')
+
+test 'Test keyboard and mouse interaction', ->
+  expect 17
+  selectedText = null
+
+  Ember.run ->
```
azirbel: do we really need `Ember.run`? I don't understand enough about testing here, but I was able to remove it in an Iverson test I wrote recently.

tests/unit/multi_select_test.coffee
```diff
@@ -0,0 +1,96 @@
...
+    validateFocusAndDropdown = (messageVisible, messageFocus) ->
+      ok isVisible(find '.ember-select-results', multiSelectComponent, messageVisible)
+      ok isFocus(textField,multiSelectComponent, messageFocus)
+
+    multiSelectComponent.focus()
+    wait().then ->
+      # test pressing ENTER key to open dropdown
+      keyEvent(multiSelectComponent, 'keydown', 13)
+    .then ->
+      ok isVisible(find '.ember-select-results', multiSelectComponent, 'Dropdown list should appear after pressing Enter')
```
azirbel: you can keep within the 80 character limit in many places just by putting a newline before the description for these checks. In our tests we usually put these on separate lines anyway.

tests/unit/multi_select_test.coffee
```diff
@@ -0,0 +1,96 @@
...
+      resultItems = find '.ember-select-result-item', multiSelectComponent
+      ok $(resultItems[0]).hasClass('highlighted'), 'The first option should be highlighted'
+      # test selecting option using ENTER key
+      selectedText = $(resultItems[0]).text().trim()
+
+      keyEvent(multiSelectComponent, 'keydown', 13)
```
azirbel: you should make a helper for these `keyEvent`s. Something like `pressEnter(multiSelectComponent)` would read better here.

tests/unit/multi_select_test.coffee
```diff
@@ -0,0 +1,96 @@
...
+    .then ->
+      validateFocusAndDropdown('Dropdown list should still be visible after selecting an option','Text field should still be focused after selecting using Enter')
+
+      # test if selected Item is actually selected
+      resultItems = find '.ember-select-search-choice', multiSelectComponent
+      currentText = $(resultItems[resultItems.length - 1]).text().trim()
+      # trim the special character "x" at the end of the select choice text
```
azirbel: why is there a special character "x"?

tests/unit/select_test.coffee
```diff
@@ -0,0 +1,120 @@
...
+  Ember.run ->
+    selectComponent = select.$()
+
+    ok isHidden(find '.ember-select-results', selectComponent, 'Dropdown list should not exist at the beginning')
+    selectComponent.focus()
+    wait().then ->
+      # test pressing ENTER key to open dropdown
+      keyEvent(selectComponent, 'keydown', 13)
+    .then ->
+      ok isVisible(find '.ember-select-results', selectComponent, 'Dropdown list should appear after pressing Enter')
```
azirbel: check that it is initially not visible

tests/unit/multi_select_test.coffee
azirbel: check that it is initially not visible

tests/unit/select_test.coffee
```diff
@@ -0,0 +1,120 @@
...
+      # test pressing DOWN arrowkey to navigate selection down
+      keyEvent(selectComponent, 'keydown', 40)
+
+    .then ->
+      resultItems = find '.ember-select-result-item', selectComponent
+      ok $(resultItems[1]).hasClass('highlighted'), 'The second option should be highlighted'
+
+      # test pressing UP arrowkey to navigate selection down
+      keyEvent(selectComponent, 'keydown', 38)
+
```
azirbel: got some extra newlines in this file. I liked your style before where you'd do the `keyEvent` (or whatever action) and the `.then` on the following line.

tests/unit/select_test.coffee
```diff
@@ -0,0 +1,120 @@
...
+      # test if selected Item is actually selected
+      resultItems = find '.ember-select-result-item', selectComponent
+      currentText = $(resultItems[0]).text()
+      equal(selectedText, currentText)
+
+      # test if dropdown appears when we start typing letter ('a' is input here)
+      keyEvent(selectComponent, 'keydown', 97)
+
+    .then ->
+      ok isVisible(find '.ember-select-results', selectComponent, 'Dropdown list should appear after pressing a letter')
```
azirbel: check that it isn't there before you start typing

azirbel:
This is great. I think my comments are fairly easy to address (but I hope you do address them; there are some old comments still there which I still think are valid). LGTMBC.

src/multi_select.coffee
thangdinh: I think it comes from `legacy` code when I was new to ember and coffee script. I saw one place having it and i was afraid of observers and such, that's why I put the check in to make sure :-)

tests/unit/multi_select_test.coffee
thangdinh: it's the close button next to selected options (pills)

tests/unit/multi_select_test.coffee
thangdinh: I don't know about it either. 
@cyril-sf @igillis do you have any idea?

tests/unit/select_test.coffee
thangdinh: Is it in line 90?

tests/unit/select_test.coffee
thangdinh: is the line 62 sufficient, @azirbel ?

thangdinh:
@azirbel , which are old comments still there which are valid? I thought I addressed all of them. Have I missed any of them? Could you please help me point out which ones you found if you still remember?

src/multi_select.coffee
azirbel: ok, it's not a big deal but if you are able to try it out with the simpler versions of code that would be awesome :+1:

src/select.coffee
azirbel: here's a comment from a while ago which still sounds valid

tests/unit/multi_select_test.coffee
azirbel: ok, please add a comment to explain that :)

tests/unit/select_test.coffee
azirbel: oh, yeah! was that there the whole time? must have missed it

tests/unit/select_test.coffee
azirbel: again, yes ... my mistake. looks good

azirbel:
Added a new reply to the "old valid" comment so you don't miss it.

src/multi_select.coffee
```diff
@@ -22,6 +22,11 @@ Ember.Widgets.MultiSelectComponent = Ember.Widgets.SelectComponent.extend
   placeholder: undefined
   persistentPlaceholder: undefined
 
+  # disable tabindex of the component container to set focus directly to
+  # the input field, which is always visible. This helps reducing one tab
+  # step to navigate back to the previous component
+  tabindex: -1
```
thangdinh: @azirbel I already had the tabindex here overridden as -1 as you mentioned in the other comment

azirbel: oh, ok - great! Since the other comment was still around, I figured you hadn't addressed it; you might want to comment "done" or something as a reply next time so I can easily tell the comment has been resolved when I skim through.

src/select.coffee
thangdinh: Done. Found as much as I can.

src/mixins.coffee
thangdinh: done

src/multi_select.coffee
thangdinh: done

src/select.coffee
thangdinh: I refactored it but right now it is not relevant (temporarily) because of the highlighted item bug, I will postpone the showing dropdown in multi-select.

src/select.coffee
thangdinh: Done

tests/helpers/test_helpers.coffee
thangdinh: Done

tests/unit/multi_select_test.coffee
thangdinh: I removed it and it ran fine.

tests/unit/multi_select_test.coffee
thangdinh: Done

tests/unit/multi_select_test.coffee
thangdinh: Done

tests/unit/multi_select_test.coffee
thangdinh: Done

tests/unit/select_test.coffee
thangdinh: Done

tests/unit/select_test.coffee
thangdinh: Done

src/multi_select.coffee
thangdinh: It's my bad. I'll do it from now on. Thanks.

Gruntfile.coffee
```diff
@@ -34,15 +34,15 @@ module.exports = (grunt) ->
       unit:
         configFile: 'karma.conf.js'
         singleRun: false
-        exclude: ['build/src/ember_widgets.js', 'tests/functional/*.js', 'tests/integration/*.js'],
+        exclude: ['build/src/ember_widgets.js', 'build/tests/functional/*.js', 'build/tests/integration/*.js'],
```
cyril-sf: If you rebase, that should be fixed on master

src/mixins.coffee
```diff
@@ -1,3 +1,16 @@
+Ember.Widgets.KeyboardHelper = Ember.Mixin.create
```
cyril-sf: Much needed :)

src/multi_select.coffee
```diff
@@ -1,4 +1,4 @@
-Ember.Widgets.MultiSelectOptionView = Ember.View.extend
```
cyril-sf: What are the chances that someone was relying on that class name?

src/multi_select.coffee
```diff
@@ -15,12 +15,24 @@ Ember.Widgets.MultiSelectOptionView = Ember.View.extend
     @notifyPropertyChange 'label'
   , 'context', 'labelPath'
 
+Ember.Widgets.MultiSelectOptionView = Ember.Widgets.SelectOptionView.extend
+  processDropDownShown: ->
+    @_super()
+    @get('controller').focusTextField()
```
cyril-sf: A controller should not manipulate the DOM, it breaks the separation of concerns

src/multi_select.coffee
```diff
@@ -15,12 +15,24 @@ Ember.Widgets.MultiSelectOptionView = Ember.View.extend
...
+
 Ember.Widgets.MultiSelectComponent = Ember.Widgets.SelectComponent.extend
   layoutName: 'multi-select'
   selections: undefined
   choicesFieldClass: ''
   placeholder: undefined
   persistentPlaceholder: undefined
+  resetQueryOnSelect: true
```
cyril-sf: `yes` for our CS lover friends

src/multi_select.coffee
cyril-sf: It seems like `controller` is actually a component.

src/select.coffee
```diff
@@ -166,8 +182,7 @@ Ember.AddeparMixins.ResizeHandlerMixin,
   # TODO(Peter): consider calling this optionViewClass?
   itemView: Ember.computed ->
     itemViewClass = @get 'itemViewClass'
-    if typeof itemViewClass is 'string'
-      return Ember.get itemViewClass
+    return Ember.get itemViewClass if typeof itemViewClass is 'string'
```
cyril-sf: Not a huge fan of this. This assumes that the string represents the global path to the class. But you didn't write it, so we can revisit that later.

src/select.coffee
```diff
@@ -346,11 +367,27 @@ Ember.AddeparMixins.ResizeHandlerMixin,
     value
   .property 'selectableOptions.[]', 'highlightedIndex'
 
+  setFocus: ->
+    activeElem = $(document.activeElement)[0]
```
cyril-sf: Why do you need to go through jQuery?

tests/unit/multi_select_test.coffee
```diff
@@ -0,0 +1,129 @@
...
+  selectedText = null
+
+  append(multiSelect)
+
+  multiSelectComponent = multiSelect.$()
+  highlightedComponent = find '.ember-select-multi', multiSelectComponent
+  textField = find '.ember-text-field', multiSelectComponent
+
+  validateDropdownVisible = (messageVisible) ->
+    ok isVisible(find '.ember-select-results', multiSelectComponent,
```
cyril-sf: Shoudln't that be `ok isVisible(), messageVisible`?

tests/unit/multi_select_test.coffee
```diff
@@ -0,0 +1,129 @@
...
+    # test selecting option using ENTER key
+    selectedText = $(resultItems[0]).text().trim()
+
+    pressEnter(multiSelectComponent)
+  .then ->
+    validateTextFieldFocus('Text field should still be focused after
+      selecting using Enter')
+    # test if selected Item is actually selected
+    resultItems = find '.ember-select-search-choice', multiSelectComponent
+    currentText = $(resultItems[resultItems.length - 1]).text().trim()
```
cyril-sf:
```cs
currentText = $(resultItems[resultItems.length - 1]).text().replace(/ x$/, '').trim()
equal(currentText, selectedText
```

tests/unit/multi_select_test.coffee
```diff
@@ -0,0 +1,129 @@
...
+    click(resultItems[0])
+  .then ->
+    resultItems = find '.ember-select-search-choice', multiSelectComponent
+    equal(resultItems.length, 2, 'There should be 2 selected items after
+      selecting one item using mouse')
+    validateTextFieldFocus('Text field should be focused after selecting
+      using mouse')
+
+    # test deleting one selected item using mouse
+    closeButtons = find '.ember-select-search-choice-close', multiSelectComponent
```
cyril-sf: You should be able to simply write 

```cs
click  '.ember-select-search-choice-close', multiSelectComponent
```

src/multi_select.coffee
thangdinh: @cyril-sf You are right. But the original name is not very well-written. We have the SelectionOptionView in select component indicating the items in the dropdown while here it is the selected items (selected pills) which has a completely different meaning.
What should we do in this case?

src/multi_select.coffee
thangdinh: Yes, it is a component.

tests/unit/multi_select_test.coffee
thangdinh: :+1:  it works and I didn't notice it. Thanks.

tests/unit/multi_select_test.coffee
thangdinh: :+1:

tests/unit/multi_select_test.coffee
thangdinh: @cyril-sf which item will it click on assuming that `find '.ember-select-search-choice-close', multiSelectComponent` returns multiple results. I guess it will click on the 1st one?

src/select.coffee
thangdinh: @cyril-sf I don't know the other ways. Would you please tell me some?

src/multi_select.coffee
thangdinh: It is in master. I'll change it.

tests/unit/select_test.coffee
```diff
@@ -0,0 +1,131 @@
...
+  validateDropdownHidden = (messageHidden) ->
+    ok isHidden(find '.ember-select-results', selectComponent),
+      messageHidden
+
+  validateFocus = (messageFocus) ->
+    ok isFocused(selectComponent,selectComponent), messageFocus
+
+  validateDropdownHidden('Dropdown list should not exist at the beginning')
+  selectComponent.focus()
+  wait().then ->
```
cyril-sf: Use `andThen` instead ( see http://guides.emberjs.com/v1.11.0/testing/test-helpers/#toc_wait-helpers )

`andThen` is the public API than can potentially do more in the future

cyril-sf: Actually, `keyEvent` is an asynchronous helper, which means it can be chained without using `then`, `wait` or `andThen`

So we would probably better writing

```cs
append()
...
andThen() ->
    validateDropdownHidden
    selectComponent.focus()

pressEnter selectComponent
andThen ->
    validateDropdownVisible

pressDown selectComponent
andThen ->
    ....
```

src/select.coffee
cyril-sf: @thangdinh This comment doesn't seem to have been addressed.

cyril-sf:
The code looks good. I haven't tested like @igillis or @azirbel though.

There are a few comments remaining.

src/select.coffee
thangdinh: @cyril-sf it's already addressed in MultiSelectComponent as suggested.

src/multi_select.coffee
```diff
@@ -15,12 +15,24 @@ Ember.Widgets.MultiSelectOptionView = Ember.View.extend
...
   choicesFieldClass: ''
   placeholder: undefined
   persistentPlaceholder: undefined
+  resetQueryOnSelect: yes
+  itemViewClass: 'Ember.Widgets.MultiSelectItemView'
+
+  # disable tabindex of the component container to set focus directly to
+  # the input field, which is always visible. This helps reducing one tab
+  # step to navigate back to the previous component
+  tabindex: -1
```
thangdinh: @cyril-sf it's addressed here.

cyril-sf: :+1:

cyril-sf:
Is this the final working code?

If so, please rebase it the way you want it to be merged.

thangdinh:
@cyril-sf I squashed and rebased. Would you please help me review it?

thangdinh:
@cyril-sf when you have time, would you please help me review it? I need LGTM for QA to process :)

src/select.coffee
```diff
@@ -107,6 +110,19 @@ Ember.AddeparMixins.ResizeHandlerMixin,
...
+  # a map of accepted keys to show dropdown when being pressed
+  # these are keys to show dropdown when being pressed
+  acceptedKeys: Ember.computed ->
+    mappedKeys = Ember.Map.create()
+    # create a set of accepted keys from 'A'..'Z', 'a'..'z', '0'..'9'
+    # and some special keys Enter, Spacebar, Up, Down
+    keySet = _.union([@KEY_CODES.ENTER, @KEY_CODES.SPACEBAR],
+      [@KEY_CODES.DOWN, @KEY_CODES.UP],[65..90],[97..122],[48..57])
+    keySet.forEach (key) ->
+      mappedKeys[key] = true
```
cyril-sf: `yes`

tests/unit/multi_select_test.coffee
```diff
@@ -32,3 +38,111 @@ test 'Query should be reset on select if resetQueryOnSelect is true', ->
...
+    resultItems = find '.ember-select-search-choice', multiSelectComponent
+    equal(resultItems.length, 2, 'There should be 2 selected items')
+
+    # test deleting using keyboard
+  pressBackspace(textField)
+  andThen ->
+    resultItems = find '.ember-select-search-choice', multiSelectComponent
+    equal(resultItems.length, 1,
+      'There should be 1 selected item after deleting one')
+    validateTextFieldFocus('Dropdown list should still be shown after
```
cyril-sf:
```
   validateTextFieldFocus(
      'Dropdown list should still be shown after deleting using keyboard',
      'Text field should be focused after pressing delete')
```

is easier to read

tests/unit/select_test.coffee
```diff
@@ -38,6 +44,78 @@ test 'Test sorted filter', ->
...
+
+  validateDropdownHidden('Dropdown list should not exist at the beginning')
+  selectComponent.focus()
+
+  # test pressing ENTER key to open dropdown
+  pressEnter(selectComponent)
+  andThen ->
+    validateDropdownVisible('Dropdown list should appear after pressing Enter')
+
+    # test pressing DOWN arrowkey to navigate selection down
```
cyril-sf: extra space not needed

cyril-sf:
Some minor comments, but LGTM!

Nice work!

tests/unit/multi_select_test.coffee
thangdinh: :+1:

thangdinh:
Just fixed the search matcher, and added tests. Could you please help me review it, @igillis?

igillis:
lgtmbc

thangdinh:
Merged

Issue #45, opened 2014-06-02
SelectOptionView does not work on iOS devices 


acechase:
I've only tested on the iOS simulator so far, but I have noticed in my own code, and on the product demo page (http://addepar.github.io/#/ember-widgets/select) that the select event does not work on iOS devices (I've been focusing on the iPad in particular). You can open the drop down menu, but can not select anything from it.

Has anybody else looked at this problem? Before trying to figure it out myself I wanted to see if anyone else has already made progress on this.

Thanks!
Andrew

PR #168, opened 2015-01-19
Allows you to select on iOS 8

mrloop:
Pressing a Select result doesn't work in ios without this fix.

https://github.com/Addepar/ember-widgets/issues/45

### DIFF
--- a/src/css/select.less
+++ b/src/css/select.less
@@ -288,6 +288,7 @@
   overflow-y: auto;
 
   .ember-select-result-item {
+    cursor: pointer;
     position: absolute;
     display: block;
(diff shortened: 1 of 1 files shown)


mrloop:
I don't think this is needed with list-view update https://github.com/emberjs/list-view/pull/189

igillis:
Cool. 

@mrloop I'm happy to merge this still if you need that functionality soon. We probably won't be upgrading list-view until the ember-cli stuff is in.

mrloop:
@igillis I reckon just leave it out, it's not a complete fix and it should be resolved when list-view is upgraded.

PR #169, opened 2015-01-19
Fix memory leaks

cyril-sf:
@Addepar/uie @igillis @addepar-andy


src/popover.coffee
```diff
@@ -51,6 +51,10 @@ Ember.Widgets.BodyEventListener,
     @set 'visibility', 'visible'
     @set 'isShowing', yes
 
+  willDestroyElement: ->
+    @$().off $.support.transition.end
+    @_super()
```
Dylnuge: Is there an important distinction between calling the super destructor first or last? I'm wondering because I notice it's last here but first in `_removeDocumentHandlers` above.

Dylnuge:
LGTM

src/popover.coffee
cyril-sf: It shouldn't.

PR #170, opened 2015-01-21
fix backdrop without fade

bjcohen:
Fixes a case where if fade is disabled, we try to call a function on undefined. Not sure how it ever worked...

@azirbel @igillis


igillis:
lgtm

PR #172, opened 2015-01-29
removed superfluous method definitions

addepar-andy:
Since the `sendConfirm` action checks to see if `confirm` is a function or not, you can't really have those things defined as functions.

Reviewers:
@igillis, @azirbel, @Dylnuge


igillis:
lgtm

PR #176, opened 2015-02-13
[Color Picker] Set hex value in text field for custom colors

bjcohen:


src/color_picker.coffee
```diff
@@ -205,6 +205,14 @@ Ember.Widgets.ColorPicker = Ember.Component.extend
       ])
     ])
 
+  setupCustomColor: (->
+    selectedColor = @get 'selectedColor'
+    for row in @get('colorRows')
+      for color in row
+        return if colorToHex(selectedColor) is color
+    @set 'customColor', @get('selectedColor')
+  ).on 'init'
```
addepar-andy: this works even without prototype extensions?

bjcohen: Yeah, that won't unless I wrap it in `Ember.observer`

src/color_picker.coffee
```diff
@@ -205,6 +205,14 @@ Ember.Widgets.ColorPicker = Ember.Component.extend
       ])
     ])
 
+  setupCustomColor: Ember.observer(->
+    selectedColor = @get 'selectedColor'
+    for row in @get('colorRows')
+      for color in row
+        return if colorToHex(selectedColor) is color
```
igillis: nit: call `colorToHex` outside the loop

src/color_picker.coffee
```diff
@@ -205,6 +205,14 @@ Ember.Widgets.ColorPicker = Ember.Component.extend
       ])
     ])
 
+  setupCustomColor: Ember.observer(->
+    selectedColor = @get 'selectedColor'
+    for row in @get('colorRows')
```
igillis: nit: can you do `return if @get('colorRows').find (row) -> selectedColor in row`

igillis:
lgtmbc

src/color_picker.coffee
```diff
@@ -205,9 +212,16 @@ Ember.Widgets.ColorPicker = Ember.Component.extend
...
+    selectedColor = colorToHex(@get('selectedColor'))
+    return if @get('colorRows').find (row) ->
+      selectedColor in row.invoke 'toLowerCase'
+    @set 'customColor', @get('selectedColor')
+  ).on 'init'
+
   setCustomColor: Ember.observer ->
     if @get('isCustomColorValid') is true
-      @set 'selectedColor', @get 'customColor'
+      @set 'selectedColor', expandHexColor(@get('customColor'))
```
bjcohen: Added some stuff here to deal with 3-digit hex codes. Basically, I think `selectedColor` should *always* be a valid, 6-digit code while `customColor` can be whatever the user inputs.

igillis: agreed

PR #177, opened 2015-02-13
Fix computed properties not observing array changes in selects

Dylnuge:
@igillis @cyril-sf @thangdinh 

Select and multi-select components had a couple of computed properties
that weren't observing array changes. Both of those properties managed
the content being displayed in the select dropdown.


addepar-andy:
lgtm

cyril-sf:
I wonder if that change wasn't done by @ebryn because of performance issues.

Did you encounter a bug related to that?

PR #180, opened 2015-02-26
Disable prototype extensions in gh_pages

igillis:
Fix issues around prototype extensions and disable them by default in gh_pages. This should help ensure that people are testing with them disabled and thus writing code that's compatible with that environment.

@Dylnuge @azirbel @bjcohen


azirbel:
LGTM

Can you document this requirement in one or all of (1) the README, (2) CONTRIBUTING.md, and (3) The home demo page?

igillis:
Updated CONTRIBUTING

PR #181, opened 2015-03-02
Fix issue with empty optionLabelPath in select component

philpee2:
When optionLabelPath is empty, content proxy was previously observing
'content.@each.', which is invalid and prevented things from properly updating.

@igillis

### DIFF
--- a/src/select.coffee
+++ b/src/select.coffee
@@ -213,6 +213,10 @@ Ember.AddeparMixins.ResizeHandlerMixin,
 
   contentProxy: Ember.computed ->
     optionLabelPath = @get('optionLabelPath')
+    if optionLabelPath is ''
+      observableString = 'content.@each'
+    else
+      observableString = "content.@each.#{optionLabelPath}"
 
     ContentProxy = Ember.ObjectProxy.extend
       _select: null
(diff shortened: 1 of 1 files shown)


igillis:
LGTM

addepar-andy:
don't you have to include build files?

igillis:
yes, you do. good catch

philpee2:
What exactly do I need to do?

sreenath-addepar:
run `grunt dist` and commit

src/select.coffee
```diff
@@ -213,6 +213,10 @@ Ember.AddeparMixins.ResizeHandlerMixin,
 
   contentProxy: Ember.computed ->
     optionLabelPath = @get('optionLabelPath')
+    if optionLabelPath is ''
```
igillis: When you're updating the `dist` files, you might also want to do `unless optionLabelPath` here to handle null/undefined. Sorry, should have caught that earlier.

philpee2:
Got it. Should I just go ahead and merge that change, or do I need to PR again?

src/select.coffee
philpee2: I don't think it's possible for `optionLabelPath` to be null or undefined. Doesn't the component have a default?

igillis: it's `''` by default but there's nothing stopping someone from setting it to one of those values

philpee2: Fair enough

PR #182, opened 2015-03-03
Handle null/undefined optionLabelPath case

philpee2:
Also includes the build files from previous commit 
@igillis @addepar-andy


igillis:
lgtm

PR #183, opened 2015-03-11
Fix color picker not sending action on custom color selection

dinahshi:
Replaces span color preview with a `ColorPickerCell` for valid colors so that a
click will trigger a send.

@igillis @Addepar/uie @hanakimis


tests/unit/color_picker.coffee
```diff
@@ -48,3 +48,16 @@ test 'Color picker converts color to hex when color is invalid', ->
   testHexConversion "foo", undefined
   testHexConversion "rgb(a, b, c)", undefined
   testHexConversion "rgb(1,2,3)", undefined
+
+test 'Custom selected color is set as custom color', ->
+  color = '#addec0'
+  colorPicker.set 'selectedColor', color
+  equal colorPicker.get('customColor'), color
+
+test 'Clicking color picker cell sets selectedColor', ->
```
dinahshi: Not sure why this test isn't passing. The click doesn't seem to be working but the selector is definitely present. Took a look with @Dylnuge and we couldn't figure out what was wrong. Any ideas @igillis @cyril-sf?

src/css/color-picker.less
```diff
@@ -150,3 +150,10 @@
   position: relative;
   z-index: 1;
 }
+
+// Eyeballed to be the same height as color preview and width when there are 4
+// colors per row.
+.color-picker-hex-input.form-control.input-sm {
+  width: 104px;
+  height: 28px;
+}
```
Dylnuge: tagging @chrnguyen @hanakimis for css review

src/color_picker.coffee
```diff
@@ -216,10 +216,14 @@ Ember.Widgets.ColorPicker = Ember.Component.extend
...
 
-  setCustomColor: Ember.observer ->
-    if @get('isCustomColorValid') is true
-      @set 'selectedColor', expandHexColor(@get('customColor'))
-  , 'customColor', 'isCustomColorValid'
+  setCustomAsSelected: Ember.observer ->
+    selectedColor = @get 'selectedColor'
+    colorRows = @get 'colorRows'
+    for row in colorRows
+      if selectedColor in row
```
Dylnuge: Is selected color guaranteed to be a six digit hex code? If not, this might not always work; for instance `'#999' !== '#999999'` but those colors are equivalent.

src/css/color-picker.less
chrnguyen: do you really need all of those classes to select this element? did .color-picker-hex-input not work?

tests/unit/color_picker.coffee
cyril-sf: Trying to take a look. It's such a pain to run tests in the library :/

src/css/color-picker.less
chrnguyen: apparently so. /shrug looks okay to me.

hanakimis: yea we need to override the values for form-control.input-sm, but there are other styles that we inherit from them.

We could re-write it to add all the styles we need, but thought this was a ittle bit easier.

src/templates/color-picker.hbs
```diff
@@ -27,11 +27,17 @@
...
       <div class="form-group form-group-sm">
         <div class="input-group">
-          <span {{bind-attr class=":input-group-addon
-            :color-picker-custom-preview isCustomColor:active"
-            style="customColorCSS"}}></span>
+          {{#if isCustomColorValid}}
+            {{view Ember.Widgets.ColorPickerCell
+              colorBinding="customColor"
+            }}
+          {{else}}
```
bjcohen: What is `customColorCSS` bound to if the custom color isn't valid?

src/templates/color-picker.hbs
```diff
@@ -27,11 +27,17 @@
     <form {{bind-attr class=":color-picker-custom-form isCustomColorValid:valid:invalid"}}>
       <div class="form-group form-group-sm">
         <div class="input-group">
-          <span {{bind-attr class=":input-group-addon
-            :color-picker-custom-preview isCustomColor:active"
-            style="customColorCSS"}}></span>
+          {{#if isCustomColorValid}}
```
bjcohen: Does the `ColorPickerCell` look different than the span? Or is the only difference that it's clickable?

src/color_picker.coffee
bjcohen: You're probably going to want to add the validity checks and `expandHexColor` back in, possibly on L238?

src/templates/color-picker.hbs
dinahshi: There's no validation check in `customColorCSS` so it will be `background-color: #{input}` regardless of what the input is.

src/color_picker.coffee
dinahshi: These comments made me realize this function and the one above it should be combined. Will push another commit soon.

src/templates/color-picker.hbs
dinahshi: I'm guessing the suggestion is to put an action on the span and duplicate the click handler in `ColorPickerCell` in `ColorPicker` like 6709ee4?

tests/unit/color_picker.coffee
igillis: I don't think we should hold this fix up for tests since they haven't been running and seem to have some more fundamental issues that need to be addressed first.

src/color_picker.coffee
```diff
@@ -210,16 +210,13 @@ Ember.Widgets.ColorPicker = Ember.Component.extend
...
+    return @set('customColor', '') if @get('colorRows').find (row) ->
       selectedColor in row.invoke 'toLowerCase'
-    @set 'customColor', @get('selectedColor')
-
-  setCustomColor: Ember.observer ->
-    if @get('isCustomColorValid') is true
-      @set 'selectedColor', expandHexColor(@get('customColor'))
-  , 'customColor', 'isCustomColorValid'
+    @set 'customColor', selectedColor
+  , 'selectedColor', 'colorRows').on('init')
```
igillis: This will not work with prototype extensions disabled

Dylnuge: I think it will because it's the `on` function of `Ember.observer` and not the `on` function Ember loads into the `function` prototype.

igillis: I just ran the branch and it died there. I understand why you would think that works though.

Dylnuge: I can't find any documentation on what Ember would use for `init` triggers with prototypes disabled then. Does `Ember.observer` just return a plain JS function?

igillis: Seems that it does?
```
var foo = Ember.observer(function(){})
undefined
foo.constructor
function Function() { [native code] }
```

You would use `Ember.on` like it is currently.

bjcohen: https://github.com/emberjs/ember.js/blob/v1.10.0/packages/ember-metal/lib/mixin.js#L824

Dylnuge: :+1:

dinahshi: So I have to split this into an observer and `Ember.on 'init'` with the same logic?

Dylnuge: I think you can wrap the two if they're just returning functions, e.g.

```javascript
Ember.on({'init', function() {Ember.observer('thing1','thing2', function() { //my function})})
```

sreenath-addepar:
```javascript
Ember.on('init', Ember.observer('thing1', 'thing2', function() {
  //do things
}))
```

Dylnuge: Doh, that's what I meant...

tests/unit/color_picker.coffee
cyril-sf: You're using the `click` event. You need an event dispatcher in your test setup.

See https://github.com/Addepar/ember-widgets/blob/master/tests/functional/multi_select_test.coffee#L10-L11

tests/unit/color_picker.coffee
```diff
@@ -48,3 +48,16 @@ test 'Color picker converts color to hex when color is invalid', ->
   testHexConversion "foo", undefined
   testHexConversion "rgb(a, b, c)", undefined
   testHexConversion "rgb(1,2,3)", undefined
+
+test 'Custom selected color is set as custom color', ->
+  color = '#addec0'
+  colorPicker.set 'selectedColor', color
```
cyril-sf: you probably want to wrap that in an Ember run loop.

src/color_picker.coffee
cyril-sf: +1 with @sreenath-addepar

cyril-sf: I always do the mistake to think only about `function` when talking about prototype extensions, but it has some other implications ( See  http://guides.emberjs.com/v1.10.0/configuring-ember/disabling-prototype-extensions ). We should be carefult about:

> If you disable prototype extension and attempt to use native arrays with things like a template's {{#each}} helper, Ember.js will have no way to detect changes to the array and the template will not update as the underlying array changes.

tests/helpers/test_helpers.coffee
```diff
@@ -71,6 +71,40 @@ _append = (app, item) ->
...
+    dropdownButton = find '.color-picker-button', element
+    click dropdownButton
+  else
+    wait()
+
+_getSelectedColor = (app) ->
+  active = find '.color-picker-dropdown .active'
+  if active.length
+    rgbVal = active.css('background-color')
+    [_, _, _, r, g, b, a, _] = /(.*?)rgb(a)?\((\d+), (\d+), (\d+)(, (\d+))?\)/.exec(rgbVal)
```
Dylnuge: What are the first three things if not rgb values?

tests/helpers/test_helpers.coffee
```diff
@@ -71,6 +71,40 @@ _append = (app, item) ->
...
+    "#" + ((1 << 24) + (r << 16) + (g << 8) + b).toString(16).slice(1)
```
Dylnuge: Do `r` `g` and `b` already come in in hex?

Dylnuge:
LGTM. The part of the test that I had comments in could use a comment though.

tests/helpers/test_helpers.coffee
dinahshi: I'm not too familiar with this code; I moved it from some Addepar test helpers. @bjcohen?

cyril-sf: http://haacked.com/archive/2009/12/29/convert-rgb-to-hex.aspx/

cyril-sf: Not sure what can go before `rgb`

bjcohen: I think Peter wrote this originally. But `a` is third (we ignore any alpha channel)

PR #185, opened 2015-03-13
Fixed computed property on arrays in Select component

thangdinh:
@Dylnuge @igillis @azirbel


igillis:
nice -- hopefully this will fix some of the wonkiness that occurs when `content` is an array of scalar values

one nitpick: could we standardize on one of `.[]` or `.@each`? I don't feel strongly, but `.@each` probably makes it easier to extend later to `.@each.foo`.

Dylnuge:
It should be `.[]` if you're not observing anything beneath it, for performance reasons.

src/select.coffee
```diff
@@ -231,7 +231,7 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
   # in the textbox
   preparedContent: Ember.computed ->
     if @get('sortLabels') then @get('sortedFilteredContent') else @get('filteredContent')
-  .property 'sortLabels', 'filteredContent', 'sortedFilteredContent'
+  .property 'sortLabels', 'filteredContent.[]', 'sortedFilteredContent.[]'
```
Dylnuge: Not your fault, but prototype extensions shouldn't be used in this repo.

src/select.coffee
```diff
@@ -350,7 +350,7 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
       (@get('groupedContent') or []).filter (item) ->
         not Ember.get(item, 'isGroupOption')
     )
-  .property 'groupedContent'
+  .property 'groupedContent.[]'
```
Dylnuge: Same thing

src/select.coffee
```diff
@@ -363,7 +363,7 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
       index = content.indexOf value
       @setHighlightedIndex index, yes
     value
-  .property 'selectableOptions', 'highlightedIndex'
+  .property 'selectableOptions.[]', 'highlightedIndex'
```
Dylnuge: And here

Dylnuge: Actually they're pretty much everywhere. @igillis how does this work properly?

igillis: Coffeescript magic
```
foo: Ember.computed ->
  console.log 'bar'
.property 'baz'
```
compiles to 
```
foo: Ember.computed(function() {
  return console.log('bar');
}).property('baz')
```

igillis: and unlike `Ember.observer`, `Ember.computed` does not return just a function.

Dylnuge: Ah! That's the part that threw me.

cyril-sf:
It seems like we're basically trying to make Ember.Select promise aware. I'm not sure if this is what we really want.

cyril-sf:
This is maybe the real culprit https://github.com/Addepar/ember-widgets/blob/thang/selectComputedProperty/src/select.coffee#L258

igillis:
@cyril-sf why do you say we're trying to make it promise aware? `PromiseArray` should act just like a normal array once it's resolved, no?

At least in the example @thangdinh was showing me, `contentProxy` was being overridden, so I don't think that's the issue.

@thangdinh maybe you can show @cyril-sf the same code you showed me and see if he has any ideas.

thangdinh:
Thanks @igillis . I emailed @cyril-sf :)

thangdinh:
Thanks @igillis for helping me writing tests.
@cyril-sf

cyril-sf:
Tests look good :+1: 

LGTM

@igillis My initial comment about promises was irrelevant ( I mixed up 2 PRs I was tagged on )

I still feel we shouldn't have to add `'filteredContent.[]'` and `'filteredContent'`. But as this change should be harmless, it's ok for me.

thangdinh:
Merged

PR #186, opened 2015-03-19
Revert "Improve usability of Select component"

dinahshi:
This reverts commit bd24484a2c1888106ed5161e517c5f7f0b8eae1c.

@igillis @thangdinh


src/select.coffee
```diff
@@ -64,10 +64,10 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
   placeholder:        undefined
   disabled: no
 
+  # we need to set tabindex so that div responds to key events
```
igillis: missed comment

igillis:
lgtmbc

src/select.coffee
dinahshi: I used git revert, I think this is from before.

thangdinh: Your revert is right. This place is wrong not because of the revert. I fixed the position that's why it was in my commit.

PR #187, opened 2015-04-02
Fixed clicking close button on selection option pills closes popover

thangdinh:
It happens when a popover contains a multi-select, clicking to delete a selected option also closes the popover.
It is because the clickHandler checks if the popover contains the event.target or not. However, at the checking time, the event.target element is already deleted out of DOM.  
@igillis @Dylnuge @cyril-sf


igillis:
This seems like a great opportunity for a functional test where you build a popover with a multiselect in it and then do this workflow.

cyril-sf:
+1 for a test

igillis:
@thangdinh is this still necessary?

thangdinh:
Sorry I don't know why but I didn't see your question you asked earlier. I think it is still necessary. The bug is still there. Should we reopen it?

igillis:
Yep, if you're still working. I thought @azirbel fixed it elsewhere.

thangdinh:
I need to write test for this but have been busy with selectUsability recently. I'll get it done soon. Thanks.

thangdinh:
@cyril-sf @igillis We have a test covered this in our product. Could we skip writing test for now here?
If yes, can you check if it's an LGTM? Thanks.

tests/unit/multi_select_test.coffee
```diff
@@ -32,3 +38,112 @@ test 'Query should be reset on select if resetQueryOnSelect is true', ->
...
+    ok isFocused(textField,multiSelectComponent), messageFocus
+
+  multiSelectComponent.focus()
+
+  # test pressing ENTER key to open dropdown
+  pressEnter(multiSelectComponent)
+  andThen ->
+    validateDropdownVisible('Dropdown list should appear after pressing
+      Enter')
+    resultItems = find '.ember-select-result-item', multiSelectComponent
```
cyril-sf: adding `multiSelectComponent` doesn't hurt but I'm not sure it's needed.

igillis:
We still need a test for this eventually, but for now let's get the fix in.

LGTM

thangdinh:
Merged

PR #188, opened 2015-04-08
Update the label when selection, path, or selection path value changes.

liuyang-li:
@azirbel @cyril-sf @Addepar/pretzel


src/select.coffee
```diff
@@ -179,9 +179,15 @@ Ember.AddeparMixins.ResizeHandlerMixin,
...
   .property 'itemView'
 
-  selectedLabel: Ember.computed ->
-    Ember.get @get('selection'), @get('optionLabelPath')
-  .property 'selection', 'optionLabelPath'
+  init: ->
+    @_super()
+    @selectionLabelPathChanges()
+
+  selectionLabelPathChanges: Ember.observer ->
```
cyril-sf: `selectionLabelPathChanges: Ember.on 'init', Ember.observer ->` and you can delete `init`

cyril-sf:
This looks good. Can you add a unit test?

azirbel:
This is a great solution! Code LGTM, I'll second the idea of a unit test. Nitpick, but can you rename `selectionLabelPathChanges` to `selectionLabelPathObserver` (which is our conventional name for observers), or maybe `selectedLabelGenerator` (since this one is special)?

src/select.coffee
liuyang-li: Done.

liuyang-li:
A unit test has been added, and the observer was renamed.

PR #189, opened 2015-04-10
Don't reset query on multi-select selection change

azirbel:
@thangdinh @liuyang-li @igillis 

Do you know why we had this line in the first place? Is it the behavior we want?

@Addepar/ui2-developers


azirbel:
@joseph-chu what do you think of this fix?

thangdinh:
I don't know why it was there in the first place. But I just saw that behavior and would like to get rid of it too. For multi select, I think we should only clear the query text whenever the dropdown is hidden. In our case here, the dropdown is always hidden between selections, which is not efficient. I am fixing it in my PR. Thanks.
LGTM

liuyang-li:
LGTM2. Is it gonna be hard to add a test?

cyril-sf:
This is part of the first commit of the repo :smile: 

+ 1 for a test

azirbel:
I will definitely write a test here; just wanted to see if anyone had objections to this change.

joseph-chu:
I'm a little suspicious of the underlying pattern that caused this to happen in the first place (a change to the DOM was made during click propagation, leading to some failure somewhere) but this fixes it as well as any other change so it looks fine to me.

azirbel:
I've added a very simple unit test here and don't plan to do more within ember-widgets. Ideally I think we should have a functional test that makes sure the selectable items aren't re-rendered when you make a selection ... but then, I also think ideally we should take a good look at the selection handling here, because it doesn't look that great to me. (Use of an observer is a big red flag).

This change fixes a bug in our main app - I'll write a more specific test there.

Let me know if you're ok with this PR as-is.

cyril-sf:
If you revert your change, does the test break?

azirbel:
Yes.

cyril-sf:
LGTM

PR #191, opened 2015-04-15
New packaging and grunt-release-it

azirbel:
I'm a bit blocked until we move files into `addon`. There are also a few significant FIXMEs in `globals.js`.


packaging/globals.js
```diff
@@ -1,115 +1,174 @@
...
-    'utils',
-    'mixins',
-    'controllers'
-  ];
+  this.outputPrefix = 'ember-widgets';
+
+  // The old global names aren't consistent: some are on Ember.Widgets, some on
+  // Ember.AddeparMixins, and some just on Ember. For backwards-compatibility
+  // we need to maintain the same old names.
+  // FIXME(igillis): Need to remove the "-component" from some file names.
```
igillis: what's the right convention to ensure we have a `-` in the name?

packaging/globals.js
```diff
@@ -1,115 +1,174 @@
...
+    'ember-widgets/components/radio-button-component': 'Ember.Widgets.RadioButtonComponent',
+    'ember-widgets/components/radio-button-group-component': 'Ember.Widgets.RadioButtonGroupComponent',
+    // FIXME(igillis): Should use only `radio-button` in the CLI version, but keep the global
+    // name which is set here.
+    'ember-widgets/components/radio-button-wrapper-component': 'Ember.Widgets.RadioButtonWrapperComponent',
+    'ember-widgets/components/select-component': 'Ember.Widgets.SelectComponent',
+    'ember-widgets/components/typeahead-component': 'Ember.Widgets.TypeaheadComponent',
+    // FIXME(igillis): Kill text editor
+    // FIXME(igillis): What is pill-select?
+    // FIXME(igillis): Do we need dom-helper?
```
igillis: these are all done

packaging/globals.js
```diff
@@ -1,115 +1,174 @@
...
-      output.push("__exports__['default'] = window.Ember.Widgets;");
-      output.push("});");
-      fs.writeFileSync(path.join(destDir, 'globals-output.js'), output.join("\n"));
+
+      // Classes to register on the application's container. We need this
+      // because we used to refer to views by their full, global name
+      // (Ember.Table.HeaderTableContainer), but now we use the view name
+      // (header-table-container). So Ember needs to know where to find those
+      // views.
+      // FIXME(azirbel/igillis): Not sure if we need this for widgets.
```
igillis: we do if we're going to stop using the global names for things

packaging/globals.js
```diff
@@ -1,115 +1,174 @@
...
+      );
+      _this.addLinesToOutput(output, [
+            "}",
+          "});",
+        "});"
+      ]);
+
+      // For backwards compatibility, set a layoutName so the component
+      // actually renders
+      // FIXME(igillis): In Ember-Table, we removed `layoutName` from the
```
igillis: need more context -- we can chat offline

tests/dummy/config/environment.js
```diff
@@ -5,8 +5,13 @@ module.exports = function(environment) {
...
     environment: environment,
     baseURL: '/',
-    locationType: 'auto',
+    // We can't use 'auto' for the location, because this breaks gh-pages.
+    // (gh-pages doesn't support URLRewrite / server configuration files, so if
+    // you try to load a route directly, you'll get a 404.)
+    locationType: 'hash',
     EmberENV: {
+      // TODO(igillis): Disable prototype extensions
+      // EXTEND_PROTOTYPES: false,
```
igillis: this is done

igillis:
I updated the ember-cli branch to pick up recent commits, some of which affect this branch.

Still haven't been able to get ember-cli/ember-cli-coffeescript to compile CS files in the `addon/` folder, but I think we should continue with things in `app/` unless someone has a good solution. Tried upgrading ember-cli and ember-cli-coffeescript but no luck.

packaging/globals.js
azirbel: I forgot that this was the reason. Never mind :)

PR #193, opened 2015-04-28
Revert "Don't reset query on multi-select selection change"

Dylnuge:
Per discussion with @yayalice @azirbel and @hanakimis, users are more likely to expect on selecting something from a multi-select that their query is reset (e.g. type "Smith" and press enter to select John Smith, and the query should reset so they can type another name). There are some subcases where this behavior might not make sense, but we'd prefer to make it the default and allow those subcases to specifically override this functionality.

@azirbel @yayalice @igillis please review/confirm.

This reverts commit 78f5fe0cedb37465ffb828efb351a5bd8210b9f8.


sreenath-addepar:
LGTM

yayalice:
should we retain the behavior for multi-checkable?

Dylnuge:
Better idea: #194 

Closing this

PR #194, opened 2015-04-28
Allow configuration for resetting query on selection for multiselects

Dylnuge:
@azirbel @yayalice @sreenath-addepar @igillis I like this better, what do you think?

Add a configuration parameter, `resetQueryOnSelect`, which determines whether or
not to reset the query when something is selected. If set to "true" then
clicking or pressing enter on a selection clears the text in a multiselect. If
set to "false" than the query is maintained even after a selection is made.
Defaults to true, which restores the default behavior from before 78f5fe0ce.

Because 78f5fe0ce is not yet released, this is a non-breaking API change.


sreenath-addepar:
LGTM

sreenath-addepar:
oh, you should add this to the docs

Dylnuge:
Docs updated, does this still look good?

azirbel:
:+1: love this

sreenath-addepar:
:shipit:

PR #195, opened 2015-05-27
Update Ember to 1.5 and add Ember QUnit

cyril-sf:
The main goal is to write tests more easily.

Ember adds `triggerEvent` ( 1.5 ) and allows to not chain test helpers ( 1.2 ).

@thangdinh @Dylnuge @azirbel @igillis @addepar-andy @bjcohen


Dylnuge:
LGTM! :heart:

tests/unit/multi_select_test.coffee
```diff
@@ -1,36 +1,34 @@
...
-
-  teardown: ->
-    Ember.run ->
-      dispatcher.destroy()
-      multiselect.destroy()
+moduleForComponent 'multi-select', '[Unit] Multi select component'
 
 test 'Test preparedContent after some options are already selected', ->
   expect 2
+  multiselect = @subject
```
thangdinh: We used to not need to create the component for every single test. As I understand, @subject is an alias of create the component. If it is, is it good?

cyril-sf: `setup` is called for every test, so we were already creating one per test.

And that's usually a good pattern to avoid test pollution.

PR #198, opened 2015-06-10
Radio Button Group actions sends the value as its argument

philpee2:
Welp, I'm an idiot. fdd83948261c7971a81f9e1cb9b0ede2c3fd9d1c should have included the selected
value as the argument to the sent action

@thangdinh @igillis @Addepar/uie


Dylnuge:
LGTM

philpee2:
I forgot that the first argument to `sendAction` has to be the action name, and that arguments come after that. Fixed.

PR #201, opened 2015-06-11
do not execute insert callback if destroying

joseph-chu:
In some cases, particularly in test, if a modal is spawned and then clicked too quickly, the modal will no longer exist by the time Ember attempts to execute the `didInsertElement` callback, resulting in an error.

@cyril-sf


joseph-chu:
Fixed commented code @Dylnuge @philpee2

philpee2:
Code still looks commented to me?

joseph-chu:
let's pretend that didn't happen

cyril-sf:
LGTM

PR #203, opened 2015-06-23
Fixed "karma test doesn't include css file"

thangdinh:
Add link to css back to karma config file
@cyril-sf @igillis


cyril-sf:
LGTM

Does it finally fix your problem?

thangdinh:
Yes it does. But I still need to run `grunt` and Ctrl-Break so that it can starts running properly after that. I think the build is not smooth.

thangdinh:
Merged it before but forgot to close

PR #204, opened 2015-06-24
Improvement when using DISABLE_ANIMATIONS=true

cyril-sf:
@maddiehorowitz @Dylnuge @thangdinh @azirbel


tests/unit/modal_test.coffee
```diff
@@ -0,0 +1,19 @@
+moduleForComponent 'modal', '[Unit] Modal component',
+  needs: [
+    'template:modal'
+    'template:modal_header'
+    'template:modal-footer'
+  ],
+  
+  teardown: ->
+    Ember.Widgets.DISABLE_ANIMATIONS = no
```
azirbel: `modal.destroy()` ?

tests/unit/modal_test.coffee
```diff
@@ -0,0 +1,19 @@
...
+
+test 'With DISABLE_ANIMATIONS=true, hide can be called several times', ->
```
azirbel: `expect 1`?

tests/unit/modal_test.coffee
```diff
@@ -0,0 +1,19 @@
...
+  Ember.Widgets.DISABLE_ANIMATIONS = yes
+  modal = @subject()
+
+  @append()
+  andThen ->
+    modal.hide()
+    modal.hide()
+    ok yes, 'It works'
```
azirbel: Maybe more verbose, repeating "hide can be called several times"?

azirbel:
LGTMBC

tests/unit/modal_test.coffee
cyril-sf: It's already destroyed in the test, but that's not a good reason.I would hope that Ember QUnit does it for us. 

I'll check that.

PR #205, opened 2015-06-24
Don't show No Results on select empty state

addepar-andy:
If select component has no content, then don't show the "No results match" result.

Reviewers:
@Addepar/ui2-developers


igillis:
lgtm

cyril-sf:
it needs a test

sicasli:
I added a view for when the content is empty. This view can be used to show a help message about why the content is empty.

src/templates/select.hbs
```diff
@@ -30,8 +30,16 @@
     {{#if isLoading}}
       <span class="ember-select-loading">Loading...</span>
     {{/if}}
-    {{#if hasNoResults}}
-      <span class="ember-select-no-results">No results match "{{query}}"</span>
+    {{#if contentIsEmpty}}
+      <div class="ember-select-empty-content">
+        {{#if emptyContentView}}
```
cyril-sf: this condition is not tested

tests/unit/select_test.coffee
```diff
@@ -197,3 +199,42 @@ test 'Test query matching', ->
...
+    optionValuePath: 'code'
+    classNames: 'select-class-name'
+
+  @append()
+
+  selectElement = select.$()
+  openDropdown(selectElement)
+  andThen ->
+    ok isPresent(emptyContentSelector, selectElement),
+      'Empty content view displayed'
```
cyril-sf: this assertion is a bit misleading as it doesn't clearly test `emptyContentView`

src/templates/select.hbs
sicasli: done

cyril-sf: Now you only tested that code path :smile: , we need to test both!

cyril-sf: You can start your test the way it was before, then set the `emptyContentView` and check after that it's rendered

sicasli: Yeah I thought it's possible to render empty view so the test might be useless. Okay updated.

tests/unit/select_test.coffee
```diff
@@ -197,3 +199,55 @@ test 'Test query matching', ->
...
+      'Empty content block displayed'
+    ok isNotPresent('.empty-content-view', selectElement),
+      'Empty content view not displayed before specified'
+    ok isNotPresent(noResultSelector, selectElement),
+      '"No result" message not displayed'
+  andThen ->
+    Ember.run ->
+      select.set 'emptyContentView', EmptyContentView
+  andThen ->
+    ok isPresent('.empty-content-view', selectElement),
```
cyril-sf: And also

cyril-sf:
LGTM after remaining comment and tests pass

tests/unit/select_test.coffee
sicasli: wait what? `emptyContentView` is inside `emptyContentSelector`

cyril-sf: I meant `empty-content-view`

sicasli: Not sure if I understand what you mean here. `emptyContentSelector` and `empty-content-view` are both present at this point.

cyril-sf: You need to make sure that setting `emptyContentView` is going to remove that view from the DOM.

sicasli: Oh okay I see what you mean

PR #209, opened 2015-07-09
Fixed bug of setting focus on select component when it's already destroyed

thangdinh:
There is some cases we destroy the component in `userDidSelect` action --> it is crashed when we try to focus back to the component after that. 
@igillis @azirbel @Dylnuge @philpee2


src/select.coffee
```diff
@@ -413,7 +413,10 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
     item = @get 'highlighted'
     @set 'selection', item unless Ember.isEmpty(item)
     @userDidSelect(item) unless Ember.isEmpty(item)
-    @$().focus()
+    # we want to set focus back the the component.
+    # However, in some use cases, the select component is closed
+    # depending on how the userDidSelect action is implemented.
+    @$()?.focus()
     # in case dropdown doesn't close
     @send 'hideDropdown' if @get('showDropdown')
```
azirbel: this is ok, even if the component is destroying?

azirbel:
LGTM, with one comment ... not sure if there are other safeguards we need to add. But this seems reasonable in itself

src/select.coffee
thangdinh: yeah because at that time `@get('showDropdown')` gives `undefined` which is ok here.

thangdinh:
It's already merged but forgot to close.

PR #211, opened 2015-07-16
Add tooltips to select and multi-select

philpee2:
By default, options in the select and multi-select have tooltips. This can be
controlled with the `showTooltip` property.

Also replaces `yes` and `no` in documentation with `true` and `false`

@Addepar/uie @igillis @thangdinh


thangdinh:
C/c: @azirbel

philpee2:
@thangdinh Alex should be getting tagged in the UIE group, no?

app/templates/ember_widgets/documentation.hbs
```diff
@@ -100,6 +100,11 @@
       <td>&#39;&#39;</td>
       <td><p>The path of the default option.</p></td>
     </tr>
+    <tr>
+      <td>showTooltip</td>
+      <td>true</td>
+      <td><p>Whether to show a tooltip when hovering over an option.</p></td>
```
hanakimis: I would check with Patricia on this wording. Don't think she has access to ember widgets, but in the meantime, might make sense to just pull her aside for advice...

Dylnuge: @pagael

It's open source so she should be able to see this even if she can't write to the repo.

pagael: Thanks, guys. Phil asked over Slack, and it looks like he's on it.

philpee2:
bump

app/templates/ember_widgets/documentation.hbs
```diff
@@ -163,7 +169,7 @@
     </tr>
     <tr>
       <td>fade</td>
-      <td>yes</td>
+      <td>true</td>
       <td><p>If this is yes then the modal fades in/out.</p></td>
```
NicholasMohr: can you change all of these to `if this is true` if you're going to use true/false

philpee2: Good catch. Fixed

NicholasMohr:
why do we have so many views/components in one file, isn't that against convention?

src/multi_select.coffee
```diff
@@ -15,6 +15,11 @@ Ember.Widgets.MultiSelectOptionView = Ember.View.extend
     @notifyPropertyChange 'label'
   , 'context', 'labelPath'
 
+Ember.Widgets.MultiSelectTooltipItemView = Ember.Widgets.SelectTooltipOptionView.extend
+  processDropDownShown: ->
+    @_super()
+    @get('controller').focusTextField()
```
NicholasMohr: why do we not need something like this in `SelectTooltip`?

philpee2:
@NicholasMohr Generally yes, but it's done all over the place in Ember Widgets

src/multi_select.coffee
philpee2: Not sure, but that's not a new change in this commit. This is just adding the same from the regular view to the tooltip view

NicholasMohr:
LGTM then

PR #212, opened 2015-08-03
Added event to bodyClick's parameters for handling special cases.

thangdinh:
Problem: Sometimes we can have a component which is added directly to the `body` instead of our app. It causes problem when users click on that component (such as bootstrap date-picker), any hosting component using the `bodyEventListener` will suffer from it by thinking they click outside of the hosting component. So the `bodyClick` is triggered and the hosting component is closed right after that.
Solution: Adding the `event` into the parameters of `bodyClick` so that we can handle those special cases if we need to.
@azirbel @igillis @Addepar/uie


thangdinh:
bump

maddiehorowitz:
LGTM - great job adding documentation!

PR #213, opened 2015-08-17
Fixes IE by changing document.contains to jQuery

jordansmith42:


src/mixins.coffee
```diff
@@ -72,7 +72,7 @@ Ember.Widgets.BodyEventListener = Ember.Mixin.create
       Ember.run =>
         if (@get('_state') or @get('state')) is 'inDOM' and Ember.isEmpty(@$().has($(event.target)))
           # check if event.target still exists in DOM
-          if document.contains(event.target)
+          if $.contains(document.body, event.target)
```
Dylnuge: Can't it be `$('body').contains`?

philpee2:
The title of this PR presumes that IE can be fixed, which is clearly wrong. Even Microsoft agrees. 

Also, tests?

jordansmith42:
@philpee2 we will be implementing Open Sauce to catch these types of issues, also we'll look into just having a linter to catch the common cases.

philpee2:
Sounds good. Although I still don't quite understand the cause of the bug. Does IE really not support `.contains`?

Dylnuge:
No, this is Chrome's fault. IE is right in this case. `document.contains` isn't a thing. It's an alias for `document.body.contains`

Dylnuge:
LGTM

philpee2:
Oh I see. That makes sense.

Dylnuge:
Actually `document.contains` should be a thing, as a document should be a fully valid node interface. Not sure why it doesn't work.

cyril-sf:
@philpee2 We are going to use Saucelabs to do cross browser testing ( https://saucelabs.com )

thangdinh:
@Dylnuge https://developer.mozilla.org/en-US/docs/Web/API/Node/contains

Dylnuge:
Wait this needs dist files.

Also @thangdinh I saw that issue but it is in fact considered a node in IE: https://msdn.microsoft.com/en-us/library/ms535862(v=vs.85).aspx. It just seems that their `Node.contains` isn't implemented for documents, for some reason?

thangdinh:
Right, we need dist files here.

jordansmith42:
@thangdinh I'm working on the dist files, github is having issues which is causing issues with the build: https://status.github.com/

jordansmith42:
@Dylnuge @thangdinh I added this dist files in this PR: https://github.com/Addepar/ember-widgets/pull/214

Dylnuge:
It looks like IE contains is implemented on the element level.

PR #215, opened 2015-08-18
Improve usability of popover 

thangdinh:
Basically, the idea is to apply the "tab inside modal" to popover:
+ Allow tab only inside popover 
+ Avoid using global key handler to handle ESC key.

@Addepar/pretzel @Addepar/uie @igillis


src/popover.coffee
```diff
@@ -45,6 +45,12 @@ Ember.Widgets.BodyEventListener,
 
   didInsertElement: ->
     @_super()
+
+    # Make sure that after the popover is rendered, set focus to the first
+    # tabbable element
+    Ember.run.schedule 'afterRender', this, ->
```
igillis: Looks like this is duplicated here and in modal. Could we move it into `TabbableModal`? Not sure in what order the `didInsertElement` hooks get run.

igillis:
lgtmbc

src/popover.coffee
thangdinh: Good point. I just refactored it.

embooglement:
@thangdinh can we close this?

embooglement:
Closing this for now. Feel free to reopen.

PR #217, opened 2015-09-04
Fix popover teardown in popover links

Dylnuge:
@sreenath-addepar @cyril-sf @igillis @philpee2 

When a popover link is destroyed, it should destroy the popover it has created,
if any. The code in place to do this didn't work because it was trying to hit
the computed property cache but the popover wasn't stored in a computed
property. The fix is to just get the popover and destroy it if it exists.


embooglement:
LGTM

src/popover_link.coffee
```diff
@@ -9,15 +9,12 @@ Ember.Widgets.PopoverLinkComponent = Ember.Component.extend
   popoverClassNames: []
   rootElement: '.ember-application'
   fade: yes
+  _popover: null
 
-  willDestroy: ->
+  willDestroyElement: ->
+    @get('_popover')?.destroy()
```
cyril-sf: no one seems to set `_popover` to null.

Can that create a problem?

Dylnuge: Yes, good catch. We should probably also check if `_popover` is already destroyed here.

PR #219, opened 2015-10-09
Fixed the bug: userSelected action is fired when we open the dropdown

thangdinh:
Problem: When we click on the button to open the dropdown. The button is also a selection item which is selected last time. So right now, it fires an action userSelected indicating that a new selection is made, which is wrong.
Solution: Check if the dropdown is currently open --> the click is on a new selection item.
@Addepar/fire @igillis @cyril-sf


philpee2:
Do we have any tests for select components?

thangdinh:
@philpee2  We do

philpee2:
Then can you add a test?

thangdinh:
What do you mean? 
1) How do we simply check if an action is fired? 
2) Do you suggest that we test that if we click on a select component, no action is fired? 
If yes, I do not agree with this. I don't think because we coded it in the wrong way, we should test that wrong behavior.

philpee2:
1. You  can use `sinon`
2. Yes. It's a change in behavior. Seems like it should have a test?

thangdinh:
I think it's not changing the behavior. It just changed the way the component fires an action for no reason. We are not going to test things that won't be broken again, I think.

philpee2:
Firing actions is part of the behavior of the component.

On Friday, October 9, 2015, thangdinh <notifications@github.com> wrote:

> I think it's not changing the behavior. It just changed the way the
> component fires an action for no reason. We are not going to test things
> that won't be broken again, I think.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Addepar/ember-widgets/pull/219#issuecomment-147020882>
> .
>


-- 

Phil Nachum
Software Engineer | Addepar <https://addepar.com/>
(917) 834-1367


1215 Terra Bella Ave.
Mountain View, CA 94043

thangdinh:
Agreed. But what I'm trying to say, it doesn't mean that we have some developer coded it up like this:
```
this.sendAction 'foo'
this.sendAction 'bar'
```
Then later, when we see they're completely useless, we remove those 2 lines. And now we have to add a test saying:
```
test 'No action fired when clicking on the component' 
```
I don't mind writing tests. I just don't think it's a good idea to write random tests.

philpee2:
Up to you I guess.

On Friday, October 9, 2015, thangdinh <notifications@github.com> wrote:

> Agreed. But what I'm trying to say, it doesn't mean that we have some
> developer coded it up like this:
>
> this.sendAction 'foo'
> this.sendAction 'bar'
>
> Then later, when we see they're completely useless, we remove those 2
> lines. And now we have to add a test saying:
>
> test 'No action fired when clicking on the component'
>
> I don't mind writing tests. I just don't think it's a good idea to write
> random tests.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Addepar/ember-widgets/pull/219#issuecomment-147021586>
> .
>


-- 

Phil Nachum
Software Engineer | Addepar <https://addepar.com/>
(917) 834-1367


1215 Terra Bella Ave.
Mountain View, CA 94043

cyril-sf:
The test should be fairly easy: no action is sent when opening the dropdown

thangdinh:
This PR is no longer valid

PR #220, opened 2015-10-13
Allow select component to have ability to always show its dropdown

thangdinh:
There are some other problems I encountered. So I figure out that we might need to have a way to indicate that we want the select component to always show its dropdown. 
Please help me review it: @Addepar/fire @igillis


embooglement:
I'm not a big fan of this API change. It seems weird to me that `showDropdown` and `alwaysShowDropdown` are different things. For instance, what happens when `alwaysShowDropdown==true` and `showDropdown==false`?

src/select.coffee
```diff
@@ -140,16 +153,39 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
       @get('originalItemViewClass')
   .property 'showTooltip'
 
+  ###*
+  * Function to be called when we want to open the dropdown, i.e. we have to
+  * clean up the query, render the dropdown...
```
embooglement: I don't think our docs should have ellipses in them, especially for open source projects. The other items in that list may be obvious to you, but not others.

thangdinh:
`showDropdown` is for the toggle of the dropdown. If it is `false`, then we you first open the select component does have not dropdown. After you open the dropdown (clicking, hitting keyboard...), it will be locked there, makes sense?

embooglement:
So, again, what happens when someone passes in `alwaysShowDropdown: true` and `showDropdown: false`?

embooglement:
To me it seems like `showDropdown` should have three valid values, `true`, `false`, and `'always'`

src/select.coffee
```diff
@@ -140,16 +153,39 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
...
+    # start typing their new query right away.
+    @set 'query', ''
+
+    @updateDropdownLayout()
+
+    # Set the focus to the search field so that users can start typing their
+    # query right away.
+    @_setSearchFieldFocus = Ember.run.schedule 'afterRender', this, ->
+      if (@get('_state') or @get('state')) is 'inDOM'
+        @$('.ember-select-search .ember-text-field:eq(0)').focus()
```
embooglement: Kind of nitpicky, but jQuery has a `:first` selector that you can use instead of `:eq(0)`

src/select.coffee
```diff
@@ -140,16 +153,39 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
...
+
+    # cancel the set focus schedule if we have it set up in openDropdown
+    Ember.run.cancel @_setSearchFieldFocus if @_setSearchFieldFocus
+
     @_super()
 
-  updateDropdownLayout: Ember.observer ->
-    return if (@get('_state') or @get('state')) isnt 'inDOM' or @get('showDropdown') is no
+  updateDropdownLayout: ->
+    return if (@get('_state') or @get('state')) isnt 'inDOM'
```
embooglement: Can we make `(@get('_state') or @get('state')) isnt 'inDOM'` into a computed property and use it throughout this file?

embooglement: Or the inverse, whatever makes more sense

src/select.coffee
```diff
@@ -140,16 +153,39 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
...
       contentProxy.destroy()
+
+    # cancel the set focus schedule if we have it set up in openDropdown
+    Ember.run.cancel @_setSearchFieldFocus if @_setSearchFieldFocus
+
     @_super()
 
-  updateDropdownLayout: Ember.observer ->
-    return if (@get('_state') or @get('state')) isnt 'inDOM' or @get('showDropdown') is no
+  updateDropdownLayout: ->
```
embooglement: REMOVING AN OBSERVER :+1: :+1: :+1:

src/select.coffee
```diff
@@ -349,11 +372,19 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
     @set 'selection', content.findProperty(defaultPath)
   , 'content.[]'
 
-  selectableOptionsDidChange: Ember.observer ->
```
embooglement: Same level of excitement here as above

embooglement:
What's the use case for this change?

thangdinh:
Both of these `alwaysShowDropdown` and `showDropdown` are for people to pass in at the time they create the component, i.e. initialize it --> my answer above is to explain the scenario you mentioned.
"showDropdown is for the toggle of the dropdown. If it is false, then we you first open the select component does have not dropdown. After you open the dropdown (clicking, hitting keyboard...), it will be locked there, makes sense?"

embooglement:
So I think this property should be renamed to `lockDropdownOnShow` or something like that.

thangdinh:
@embooglement I gave a second thought to this feature and I think I'm going to fix the issue where we allow dropdown to be shown at the beginning. That will be enough. I will make a new PR for that and some observers fixed.

thangdinh:
I will close this one for now

PR #221, opened 2015-10-13
Add ability for dropdown to be shown at the beginning

thangdinh:
There are some use cases when we want to have dropdown shown by default, it's not currently supported. 
Tests are added.
@Addepar/fire


src/select.coffee
```diff
@@ -349,11 +368,19 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
...
+  * item does not exist.
+  * @function
+  ###
+  updateHighlightedItem: ->
     if @get('showDropdown')
       highlighted = @get('highlighted')
       if not @get('selectableOptions').contains(highlighted)
         @set 'highlighted', @get('selectableOptions.firstObject')
+
+  selectableOptionsDidChange: Ember.observer ->
```
philpee2: :cry:

tests/unit/select_test.coffee
```diff
@@ -116,6 +116,62 @@ test 'Test keyboard interaction', ->
...
+  expect 3
+
+  select = @subject
+    content: ['foo', 'bar', 'barca', 'baz']
+  @append()
+
+  selectComponent = select.$()
+
+  validateDropdownVisible = (messageVisible) ->
+    ok isVisible(find '.ember-select-results', selectComponent),
```
philpee2: Is `find` needed here?

thangdinh:
retest

src/select.coffee
```diff
@@ -80,8 +81,19 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
...
+  * @type {Boolean}
+  ###
   showDropdown: no
 
+  ###*
+  * Flag indicating that dropdown will always appear as part of the select
+  * component
+  * @type {Boolean}
+  ###
+  lockDropdownOnShow: no
```
jordansmith42: Can we prefix this with `is` or `should`? Kind of a weird name too, shouldInitWithDropdown or something along those lines?

src/select.coffee
jordansmith42: Why is this necessary?

src/select.coffee
```diff
@@ -492,8 +531,10 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
 
   actions:
     toggleDropdown: (event) ->
-      return if @get('disabled')
-      @toggleProperty 'showDropdown'
+      if @get 'showDropdown'
+        @send 'hideDropdown' if not @get('lockDropdownOnShow')
```
jordansmith42: Why handle this differently than other cases of showDropdown/lockDropdownOnShow?

src/select.coffee
embooglement: I like Jordan's suggestio

src/select.coffee
```diff
@@ -42,11 +43,11 @@ Ember.Widgets.SelectOptionView = Ember.ListItemView.extend
   click: ->
     return if @get('content.isGroupOption')
     @set 'controller.selection', @get('content')
-    @get('controller').userDidSelect @get 'content'
-    # if there's a selection and the dropdown is unexpanded, we want to
-    # propagate the click event
-    # if the dropdown is expanded and we select something, don't propagate
+    # Check if the dropdown is currently shown to make sure that we has clicked
```
embooglement: Should be "make sure that we have clicked"

src/select.coffee
```diff
@@ -42,11 +43,11 @@ Ember.Widgets.SelectOptionView = Ember.ListItemView.extend
   click: ->
```
embooglement: Should the comments in this function be included as documentation instead?

src/select.coffee
```diff
@@ -80,8 +81,19 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
   # we need to set tabindex so that div responds to key events
   tabindex: 0
 
+  ###*
+  * Flag controlling the dropdown's appearance of the select component
```
embooglement: This comment is a little confusing to me. I think it should be something like "Flag that determines whether or not the dropdown is shown".

src/select.coffee
thangdinh: It's not `init with Dropdown` but instead, it will never collapse the dropdown. 
So Jordan+Kevin's suggestions will become `shouldLockDropdownOnShow`? And yeah, we have 80 characters length limit.

embooglement: `shouldLockDropdownOnShow` seems fine to me.

src/select.coffee
thangdinh: 1. I didn't put an attempt to fix all of the observers in this component in this PR. I happened to find some of the observers are straightforward to be remove so I remove them. 
2. The reason of this observer (as part of the original solution) was to tackle the problem that the content of the selection is changed, for which the current highlighted item does not exist anymore. 
It is not straightforward to build computed properties because here we need the highlighted index bound to the UI, while we need the highlighted item for other operations. With only highlighted index, we cannot know if the item is already removed or not. Something like that. I can explain in person when we want to remove this observer.

src/select.coffee
thangdinh: What do you mean? 
This is for the case of toggling the dropdown. Where we have an `else` case to open the dropdown if it's not there yet. Other logics are aiming to hide the dropdown only.
Is it the answer of your question?

tests/unit/select_test.coffee
thangdinh: It was there before we shaped the way we write tests. I will remove all of them when I refactor the tests here.

src/select.coffee
jordansmith42: Works for me

src/select.coffee
jordansmith42: Yep, I got it now

thangdinh:
I decided to close this one again as I have a simpler solution for my problem https://github.com/Addepar/ember-widgets/pull/222.
I will revise this PR as an improvement PR (killing observers, and newly added tests) later.

PR #222, opened 2015-10-14
Prevent userSelected action to be called when the drop down is not shown

thangdinh:
Problem: `userDidSelect` is called even when the dropdown is not shown. The reason is we use the same result item inside the toggle button and clicking on it is equivalent to clicking on one item in the dropdown, which then triggers the `userDidSelect`.
Solution: Override the `click` function inside the `selectedItemView` which is the item displayed inside the toggle button.

Paired with @philpee2 . Please help me review this PR, @embooglement @Addepar/fire


src/select.coffee
```diff
@@ -209,6 +209,13 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
       tagName: 'span'
       labelPath: Ember.computed.alias 'controller.optionLabelPath'
       context: Ember.computed.alias 'controller.selection'
+      ###*
+      * This is to override the click event in the SelectOptionView
+      * so that we can make sure only when we click on a new item in the list,
+      * the userDidSelect action will be fired, not when we open the dropdown.
+      * @override
+      ###
+      click: -> Ember.K()
```
embooglement: `click: Ember.K`

src/select.coffee
```diff
@@ -209,6 +209,13 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
       tagName: 'span'
       labelPath: Ember.computed.alias 'controller.optionLabelPath'
       context: Ember.computed.alias 'controller.selection'
+      ###*
+      * This is to override the click event in the SelectOptionView
+      * so that we can make sure only when we click on a new item in the list,
+      * the userDidSelect action will be fired, not when we open the dropdown.
```
philpee2: I would add some clarification here that this view for the selected item is an extension of the view used to display each option in the popover list, and we don't want the click handler for the options in the popover list to be triggered when you click on the selected item. It's not really about the `userDidSelect` action. That just happens to be one of the effects of the click handler

cyril-sf:
we need to figure out why the tests are failing.

philpee2:
Hm, they were passing locally. Is that an old run?

cyril-sf:
I don't think so

embooglement:
They are failing locally when running the entire module, so this is probably a test pollution problem.

src/select.coffee
```diff
@@ -209,6 +209,17 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
       tagName: 'span'
       labelPath: Ember.computed.alias 'controller.optionLabelPath'
       context: Ember.computed.alias 'controller.selection'
+      ###*
+      * Note: This view is an extension of the view used to display used to
```
philpee2: `used to display used to display`

src/select.coffee
```diff
@@ -209,6 +209,17 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
...
+      * display each option in the dropdown list.
+      * We would like to override the click event in the SelectOptionView
+      * so that we can make sure that only when we click on a new option in the
+      * list, the click event in the SelectOptionView is fired, not when we open
+      * the dropdown. It will allow us to trigger the change item event when we
+      * click to select a new option in the dropdown.
+      * @override
+      ###
+      click: -> Ember.K
```
philpee2: Can this need to be `click: Ember.K`?

tests/unit/select_test.coffee
```diff
@@ -122,6 +121,32 @@ test 'Test keyboard interaction', ->
       pressing ESC')
     validateFocus('Select component should be focused after pressing ESC')
 
+test 'Test userSelected action', ->
+  expect 3
+
+  select = @subject
```
philpee2: Does any of this need to be torn down to prevent test pollution?

src/select.coffee
jordansmith42: It's a noop http://emberjs.com/api/#method_K

src/select.coffee
thangdinh: Done

src/select.coffee
thangdinh: I think we are always use `click: Ember.K`?

tests/unit/select_test.coffee
jordansmith42: Looks like @cyril-sf already merged his fix for the test pollution

PR #223, opened 2015-10-15
Fix test pollution by tearing down components properly

cyril-sf:


tests/unit/multi_select_test.coffee
```diff
@@ -5,39 +7,43 @@ moduleForComponent 'multi-select', '[Unit] Multi select component',
     'template:select-item-layout'
     'template:select-item'
   ]
+  teardown: ->
+    Ember.run ->
+      multiSelect.destroy()
+    multiSelect = null
```
embooglement: How does this work? Wouldn't `multiSelect` be `null` when the run loop is called?

embooglement:
How exactly does this fix things?

tests/unit/multi_select_test.coffee
jordansmith42: Yeah I agree this seems strange that it wouldn't be inside the ember.run

cyril-sf: `Ember.run` doesn't schedule anything, it executes the function inside a run loop.

So the execution order isn't changed.

embooglement: Ah, I see. This seems fine then.

embooglement:
LGTM

cyril-sf:
The component is not torn down properly and an event handler remains set. Later, a test triggers the event, but the component doesn't exist anymore.

This makes sure that each instantiated components is torn down, which will remove the event handler.

PR #225, opened 2015-10-22
Send action out from carouselComponent 

thangdinh:
Right now when we have a parentView containing this component, we have to observe the transition ended, while we could have an action to signal this event.
@Addepar/fire @igillis


src/carousel.coffee
```diff
@@ -81,6 +81,11 @@ Ember.Widgets.CarouselComponent = Ember.Component.extend
         $active.removeClass(['active', direction].join(' '))
         @set 'sliding', no
 
+        # Send an action out to signal that the sliding transition has ended
+        # It can is useful to avoid having observers in any parentView having
+        # this component.
```
embooglement: I don't think this explanation is really all that necessary. It's pretty much just a description of what actions in Ember do.

src/carousel.coffee
```diff
@@ -81,6 +81,11 @@ Ember.Widgets.CarouselComponent = Ember.Component.extend
...
+        @send 'transitionEnded'
```
jordansmith42: I don't think you need the observer comment as it is not specific to the implementation, more of a general concept around observers/actions.

jordansmith42:
Tests?

src/carousel.coffee
thangdinh: So what do you suggest to put in when we add an action to our code without giving any context? 
I'm not sure but I think general concept around observers/actions may be irrelevant to people who knew it, but not necessary to be useless for people who don't.

jordansmith42: I agree, but I don't think documenting ember concepts around individual actions within a component library is necessary. Documentation about ember concepts should be part of the ember documentation, documentation about ember widgets should be part of this repo.

embooglement: When we add any actions, we should document when the action is triggered, and what data it is triggered with. It is part of the public API of this component. 

Our comments should assume that the people looking at the source have a basic literacy of Ember and it's concepts.

thangdinh: Well, so is the suggestion to remove the comments?

thangdinh:
I don't know how to initiate the unit test for this carousel component

thangdinh:
Test is added. Thanks @cyril-sf .

embooglement:
LGTM

PR #226, opened 2015-11-07
Allow setting enterToConfirm

thangdinh:
Right now, we do not allow user to press Enter to confirm the modal. Add this option so that we can turn it on when needed.
@Addepar/fire


tests/unit/modal_test.coffee
```diff
@@ -74,3 +74,16 @@ test 'Test preserving the focus when clicking on non-focusable element', ->
...
+
+  modal = @subject
+    enterToConfirm: yes
+  spy = sinon.spy modal, "send"
+
+  @append()
+  modalComponent = modal.$()
+  keyEvent(modalComponent, 'keydown', 13)
+  andThen ->
+    ok spy.calledOnce, "sendConfirm gets called when hitting enter"
```
philpee2: Can you use `calledWithArguments` or whatever it's called to make this more specific?

philpee2:
What if I have focus on an inner element of the modal which is supposed to handle the `enter` key in its own way?

thangdinh:
It's turned off by default. If we want to use this feature, turn it on. If we have focus on the inner element and want to handle Enter key there while still wanting this feature, stop the propagation.

tests/unit/modal_test.coffee
thangdinh: Done

embooglement:
LGTM

PR #227, opened 2015-11-13
Fix the Ember.K misplaced

thangdinh:
I accidentally added the Ember.K for the `confirm`, `cancel`, and `close` callbacks. It essentially broke the flow of sending actions out to `targetObject`.
This PR is to remove those lines
@jordansmith42 @Addepar/fire


jordansmith42:
LGTM

PR #228, opened 2015-12-08
[Select] Fixes for Select `content` without prototype extensions

bjcohen:
Fix for `select` when `content` doesn't have prototype extensions.

@Addepar/fire @addepar-andy


philpee2:
LGTM

PR #229, opened 2016-01-05
Make color picker easier to customize and improve its rendering

thangdinh:
The current color picker is hard to customize the style without copying and modifying the whole template file. This PR is to break the color picker button (the one when you already selected a color) and the dropdown picker, which has a lot of color cells. 
Also, the current implementation always renders the dropdown and just hide it using css. It is not practical when we have to render a large amount of color pickers, for e.g. when we need to set colors for each item in a list. This PR is to only render the dropdown when we open it (click on the button). 

I need some review/discussion of best practice before starting writing tests for it.
@Addepar/fire


thangdinh:
Close this one for a different approach.

PR #230, opened 2016-01-15
ColorPicker improvement

thangdinh:
The current color picker is hard to customize the style without copying and modifying the whole template file. This PR is to break down the color picker component into sub-components the color picker button (the one when you already selected a color) and the dropdown picker, which has a lot of color cells. 
Also, the current implementation always renders the dropdown and just hide it using css. It is not practical when we have to render a large amount of color pickers, for e.g. when we need to set colors for each item in a list. In this PR we only render the dropdown when we open it (click on the button).
Other things done: Refactor code, split files, replace view by component, give some love on the introduction site.
@Addepar/fire 
CSS/website change: @Addepar/design


src/color_picker.coffee
```diff
@@ -1,258 +1,84 @@
...
-      @set 'selectedColor', color
-
-    sendCustomColor: ->
-      color = @get 'customColor'
+    ###*
+     * This action is bound to the colorPicker button to hide/show the dropdown
+     * when users click on it.
+    ###
+    toggleDropdown: ->
+      shouldOpenDropdown = @get 'shouldOpenDropdown'
```
philpee2: Can this be `@toggleProperty('shouldOpenDropdown')`?

src/color_picker_cell.coffee
```diff
@@ -0,0 +1,25 @@
...
+ * @class
+ * @augments {Ember.View, Ember.Widgets.StyleBindingMixin,
+ *            Ember.Widgets.ColorPickerMixin}
+###
+Ember.Widgets.ColorPickerCellComponent =
+Ember.Component.extend Ember.Widgets.StyleBindingsMixin,
+Ember.Widgets.ColorPickerMixin,
+  layoutName: 'color-picker-cell'
+  classNames: ['pull-left', 'color-picker-cell']
+  classNameBindings: Ember.A ['isActive:active:inactive']
```
philpee2: Does this need to be wrapped in `Ember.A`?

src/color_picker_cell.coffee
```diff
@@ -0,0 +1,25 @@
...
+  styleBindings:  'color:background-color'
+
+  color: null
+
+  isActive: Ember.computed ->
+    @colorToHex(@get('controller.selectedColor')) is @colorToHex(@get('color'))
+  .property 'controller.selectedColor', 'color'
+
+  click: (event) ->
```
philpee2: `event` isn't being used

src/color_picker_dropdown.coffee
```diff
@@ -0,0 +1,106 @@
...
+Ember.Component.extend Ember.Widgets.BodyEventListener,
+Ember.Widgets.ColorPickerMixin,
+  layoutName: 'color-picker-dropdown'
+  dropdownClass: null
+
+  selectedColor: ''
+
+  customColor: ''
+
+  colorRows:
```
philpee2: Should this be wrapped in `Ember.computed`?

src/color_picker.coffee
```diff
@@ -1,258 +1,84 @@
...
+  ###
+  colorPickerButtonPartial: 'color-picker-button-partial'
+
+  ###*
+   * This is a boolean to control if we should render the colorPicker dropdown
+   * or not. Instead of hiding it using CSS, we use this flag to control the
+   * rendering.
+   * @type {Boolean}
+  ###
+  shouldOpenDropdown: false
```
philpee2: I think something like `isDropdownOpen` is clearer

src/color_picker_cell.coffee
cyril-sf: no

src/color_picker_dropdown.coffee
cyril-sf: yes

src/color_picker_dropdown.coffee
```diff
@@ -0,0 +1,106 @@
...
+        'transparent'
+      ])
+    ])
+
+  isColorTransparent: Ember.computed.equal 'selectedColorRGB', 'transparent'
+
+  setCustomColor: Ember.on 'init', Ember.observer ->
+    selectedColor = @get 'selectedColor'
+    selectedColor = @colorToHex(selectedColor)
+    return @set('customColor', '') if @get('colorRows').find (row) ->
```
cyril-sf: this looks weird to return a specific value in an observer.

philpee2: It's also weird to return the result of `set`, imo

thangdinh:
Most of the comments are part of the old code, I rearrange them only. I will fix them still.

src/color_picker_dropdown.coffee
```diff
@@ -0,0 +1,106 @@
...
+  formattedCustomColor: Ember.computed ->
+    customColor = @get('customColor').trim()
+    if customColor.charAt(0) isnt '#'
+      customColor = '#' + customColor
+    return customColor
+  .property 'customColor'
+
+  isCustomColorValid: Ember.computed ->
+    /(^#[0-9A-F]{6}$)|(^#[0-9A-F]{3}$)/i.test("#{@get('formattedCustomColor')}")
+  .property 'customColor'
```
PhyFrancis: `.property 'formattedCustomColor'`

src/color_picker_cell.coffee
```diff
@@ -0,0 +1,25 @@
+###*
+ * ColorPickerCell
+ *
+ * This is an internal-use view to control the logic of a color cell in
+ * the color picker.
+ * @class
+ * @augments {Ember.View, Ember.Widgets.StyleBindingMixin,
+ *            Ember.Widgets.ColorPickerMixin}
+###
+Ember.Widgets.ColorPickerCellComponent =
```
PhyFrancis: it's good to also comment on what needs to be passed in, in order to use this component, e.g. `color`

src/color_picker.coffee
```diff
@@ -1,258 +1,84 @@
...
-      return "transparent"
-    return rgbToHex(red, green, blue)
-  return undefined
-
-Ember.Widgets.ColorPickerComponent = Ember.Component.extend
+###*
+ * ColorPickerComponent
+ *
+ * This is the whole color picker component. It is the interface connecting with
+ * users.
```
embooglement: This doc doesn't really make sense to me.

src/color_picker.coffee
```diff
@@ -1,258 +1,84 @@
...
+ * @class
+ * @augments Ember.Component
```
embooglement: Should also include `Ember.Widgets.ColorPickerMixin`

src/color_picker_dropdown.coffee
```diff
@@ -0,0 +1,106 @@
...
+  ###
+  formattedCustomColor: Ember.computed ->
+    customColor = @get('customColor').trim()
+    if customColor.charAt(0) isnt '#'
+      customColor = '#' + customColor
+    return customColor
+  .property 'customColor'
+
+  isCustomColorValid: Ember.computed ->
+    /(^#[0-9A-F]{6}$)|(^#[0-9A-F]{3}$)/i.test("#{@get('formattedCustomColor')}")
```
embooglement: No need to use string interpolation on `@get('formattedCustomColor')`

src/color_picker_mixin.coffee
```diff
@@ -0,0 +1,178 @@
...
+    violet: "#ee82ee"
+    wheat: "#f5deb3"
+    white: "#ffffff"
+    whitesmoke: "#f5f5f5"
+    yellow: "#ffff00"
+    yellowgreen: "#9acd32"
+
+  ###*
+   * Convert RGB color to Hex string
+   * @param  {Number} r Red channel
```
embooglement: Probably handy to specify that the numbers have to be between 0 and 255.

tests/unit/color_picker_test.coffee
```diff
@@ -11,65 +29,119 @@ testHexConversion = (colorPicker, color, hex) ->
 
 test 'Color picker converts color to hex when color is undefined', ->
   expect 3
+  colorPicker = @subject()
   [undefined, null, 0].forEach (color) =>
```
embooglement: This doesn't need to be a fat arrow function anymore.

src/color_picker.coffee
thangdinh: Right. Forgot about that function. Thanks.

src/color_picker_cell.coffee
thangdinh: It existed there before. Changed it.

src/color_picker_cell.coffee
thangdinh: Existing code. Changed it.

src/color_picker_dropdown.coffee
thangdinh: Existing code. I didn't intend to rewrite everything :D

src/color_picker.coffee
thangdinh: isDropdownOpen sounds like a check, shouldOpenDropdown is an order (for rendering). But I don't have strong opinion here.

src/color_picker_dropdown.coffee
thangdinh: This was there before. I agree that it's weird here, but I would prefer to spend time removing this entire observer separately.

src/color_picker_dropdown.coffee
thangdinh: Thanks. Nice catch!

src/color_picker.coffee
thangdinh: Any suggestion on this?

src/color_picker.coffee
thangdinh: Sure. I forgot to update the doc after adding this mixin.

src/color_picker_dropdown.coffee
thangdinh: Done

src/color_picker_mixin.coffee
thangdinh: Done.

tests/unit/color_picker_test.coffee
thangdinh: Done

src/color_picker.coffee
embooglement: So this is the sentence that doesn't make sense to me: `It is the interface connecting with users`. What are you trying to communicate here?

thangdinh: I think it wasn't good at the beginning. I would like to emphasize that it is the component but it does have sub-components where most of the logics are in the dropdown component.

src/color_picker_cell.coffee
thangdinh: I think we can pass in any attributes belonging to the component, right? I guess we should put comment on the attributes inside the component. This code was moved from existing code (splitting the file) that's why there is no comment here.

app/templates/ember_widgets/color_picker.hbs
```diff
@@ -6,24 +6,46 @@
...
+          &#123;&#123;color-picker
+            class="btn btn-default"
+            dropdownClass="pull-left"
+          &#125;&#125;</pre>
+      </div>
+    </div>
 
     <div class="col-md-3">
-      {{view Ember.Widgets.ColorPicker class="btn btn-default" dropdownClass="pull-right"}}
+      {{color-picker class="btn btn-default" dropdownClass="pull-right"}}
```
cyril-sf: is there a reason why you let the `class` in that case and not in the previous example?

app/templates/ember_widgets/color_picker.hbs
```diff
@@ -6,24 +6,46 @@
...
     <div class="col-md-3">
-      {{view Ember.Widgets.ColorPicker class="btn btn-default"}}
+      {{color-picker}}
       <button type="button" class="btn btn-default">Comparison</button>
     </div>
+    <div class="col-md-9">
+      <div class="highlight">
+        <pre class="prettyprint lang-html">
+          &#123;&#123;color-picker
+            class="btn btn-default"
```
cyril-sf: those are not the parameters passed to the example.

app/templates/ember_widgets/color_picker.hbs
thangdinh: Now looking back these "examples" I don't think it make any sense with the button Comparison next to it. I think what we wanted before is: "Hey that's what it looks like, this is what the default button looks like". Right now the class doesn't do anything, I think.

thangdinh:
Do we have any other comment?

src/templates/color-picker-button-partial.hbs
```diff
@@ -0,0 +1,14 @@
...
+  {{#if isColorTransparent}}
+    {{color-picker-cell
+      color=selectedColor
+      classNames="color-picker-cell-transparent"
+    }}
+  {{else}}
+    {{color-picker-cell
+      color=selectedColor
+    }}
+  {{/if}}
```
embooglement: Can't this be 
```
{{color-picker-cell
  color=selectedColor
  classNames="isColorTransparent:color-picker-cell-transparent"
}}
```

embooglement:
LGTM

src/templates/color-picker-button-partial.hbs
thangdinh: Good catch! I just moved around the content didn't pay attention to this.

PR #235, opened 2016-02-18
Fix problem with using Ember Array wrappers as content in Select component

thangdinh:
### Problem
Currently we are using lodash to find the selection among the content in the select component:
`selection = _.find(content, valuePath, value)`
It does not work for array wrappers such as `ArrayProxy`, `ArrayController`, `DS.ManyArray`, for which we have to use `findProperty` (old version of `findBy` because of our current version of Ember here).

### Solution
Check if the content is a wrapper or not, if it is use `findProperty`, otherwise use `_.find`

@cyril-sf @bjcohen @Addepar/fire


tests/unit/select_test.coffee
```diff
@@ -258,3 +258,70 @@ test "Show no-result message if has content but filtered content is empty", ->
       'Empty content block not displayed'
     ok isPresent(noResultSelector, selectElement),
       '"No result" message displayed'
+
+test 'optionValuePath with POJOs', ->
+  expect 1
+
+  obj1 = {name: 'reddit', value: 1}
+  obj2 = {name: 'red', value: 2}
+  data = [obj1, obj2]
```
bjcohen: Prototype extensions are off in these tests?

bjcohen:
Is the case
```
obj1 = {name: 'reddit', value: {subvalue: '1'}}
obj2 = {name: 'red', value: {subvalue: '2'}}
data = [obj1, obj2]
select = @subject
  content: data
  optionLabelPath: 'name'
  optionValuePath: 'value.subvalue'
```

expected to work? The way it's written right now I don't think it does.

tests/unit/select_test.coffee
```diff
@@ -258,3 +258,70 @@ test "Show no-result message if has content but filtered content is empty", ->
...
+  Ember.run ->
+    select.set 'value', 2
+
+  equal obj2, select.get('selection'),
+    'The right selection is retrieved'
+
+test 'optionValuePath with Ember Objects', ->
+  expect 1
+
+  Klass = Ember.Object.extend
```
jordansmith42: I like Klass, but might just want to call it ObjectStub or something

cyril-sf: I wrote that, used to do that in ruby.

tests/unit/select_test.coffee
cyril-sf: no. This test was initially written to reproduce the problem, before I understood what the problem is.

But that's a good point and we should test that.

thangdinh:
@bjcohen I think it works in the current code

thangdinh:
I just added the test for that

bjcohen:
It works with prototype extensions off?

tests/unit/select_test.coffee
```diff
@@ -1,6 +1,7 @@
 select = null
 emptyContentSelector = '.ember-select-empty-content'
 noResultSelector = '.ember-select-no-results'
+hasExtendedPrototype = null
```
cyril-sf: unused

thangdinh:
Is there any other concern? Should I change `Klass` @cyril-sf @jordansmith42 ? :)

cyril-sf:
@bjcohen we ran the test manually with the prototype extension off.

We'll make sure that both cases are tested once we have converted the library to an Ember-CLI addon.

cyril-sf:
LGTM

@thangdinh did you test it in Iverson?

PR #236, opened 2016-03-30
Disable custom color form on submit for color picker

billylittlefield:
**Problem:**
When using the color picker to enter a custom color hex value, hitting enter would cause a full page refresh and not actually update the color. The `setCustomColor` action was never even being triggered. If a user clicked on the custom color preview tile instead of hitting enter, the change would work as expected. The root cause was due to the default form submission not being prevented/disabled/overridden.

**Fix:**
Move the `setCustomColor` action logic from the `input` field to the `form` itself, and set it to occur on `submit`.

The existing tests for color picker tested the custom color form, but would complete the form by clicking on the preview tile rather than hitting enter / submitting the form. This worked because the preview `<span>` also has an action attached to it, and this doesn't trigger form submission. I've added a test to select a custom color with form submission instead.

ADPR-13078


billylittlefield:
Ready for review @Addepar/fire

thangdinh:
Thanks for the fixing this. However, I am not sure if this is a correct fix. IMHO, we should disable the default submit on the form because this is not a form. And we should only trigger the enter key when the input is on focus, which, is arguably the intention of the user to input a custom value. After all, popover shouldn't have submit feature, I think.

billylittlefield:
Doesn't this override the default submit? The other fix I had earlier was maybe more inline with what you were saying -- we could add `onsubmit="return false"` to the `<form>` tag, then leave the `on="insert-newline` action on the `<input>` field as it was.

thangdinh:
@Addepar/design What do you think we should do with `Enter` pressing for custom color text field? Should the Enter key (to accept custom color) be only applied when the text field is on focus or is it a global action, i.e. when we hit enter, we accept that custom color no matter where the focus is?

billylittlefield:
The only way someone could lose focus on the input field is if they click on the popover in the space between the color tiles. Clicking outside the popover will close the popover, and clicking any of the color tiles will also close the popover. I think its reasonable to require the input to be in focus, as is currently the case.

Whether or not this needs to be a form at all is another question, but maybe outside of the scope of this bugfix.

hanakimis:
@thangdinh feels like when we have focus, we accept.

tests/unit/color_picker_test.coffee
```diff
@@ -130,3 +130,15 @@ test 'Click outside of the component should close the dropdown', ->
...
+
+test 'Submitting custom color form updates color and does not reload page', ->
+  colorPicker = @subject()
+  customColor = "#abc123"
+  @append()
+
+  openColorChooser()
+  fillInCustomColor(customColor)
+  andThen ->
+    $(".color-picker-custom-form").submit()
```
thangdinh: add comment here explain why you need to do this

thangdinh:
LGTM

PR #237, opened 2016-05-20
@sourceURL is deprecated

addepar-andy:
reviewers: @embooglement, @cyril-sf


Gruntfile.coffee
```diff
@@ -154,6 +154,13 @@ module.exports = (grunt) ->
           from: /The current version is .*\..*\..*./
           to: "The current version is <%=pkg.version%>."
         }]
+      sourceMap:
+        src: ['dist/js/ember-widgets.min.js', 'dist/js/ember-widgets.js'],
+        overwrite: true,
+        replacements: [{
+          from: '//@ sourceURL',
```
embooglement: What actually introduces the source map url like this? Seems like we should fix that specifically instead of just remapping stuff.

addepar-andy: `grunt-neuter` a package that hasn't been updated in 3 years

embooglement: Had a feeling it would be something like that.

embooglement:
LGTM

cyril-sf:
I know Trek who created neuter. We can always open a PR and fix this upstream and get it merged.

addepar-andy:
@cyril-sf: you tell me what we should do, and i'll do it

cyril-sf:
What you're doing is hacky, but I'm fine w/ it as we are not going to keep `neuter` forever (and it won't be hard to change if needed)

embooglement:
@cyril-sf just tested this and it works. Merging

PR #238, opened 2016-06-09
Add hideOthers option to popover

mgreenbaum:
Allow users to disable the hideAll that's triggered when
launching a popover. This change is necessary for keeping
multiple popovers open at the same time.

@billy-addepar


billy-addepar:
@jordansmith42 @cyril-sf @thangdinh

src/popover.coffee
```diff
@@ -17,6 +17,11 @@ Ember.Widgets.BodyEventListener,
   visibility: 'hidden'
   debounceTime: 100
 
+  ###*
+   * Whether launching this popover should hide other popovers on the screen
+  ###
+  hideOthers: yes
```
thangdinh: The default behavior should be `false` I guess

mgreenbaum: `true` maintains the existing behavior

thangdinh:
Please write test for this

src/popover.coffee
thangdinh: oh ok

mgreenbaum:
from https://github.com/Addepar/ember-widgets/blob/master/CONTRIBUTING.md: 

`For now, pull requests aren't required to have tests. Once we've established a testing infrastructure, pull requests that introduce new functionality should come with tests for that functionality.`

thangdinh:
It is in 2014, we just didn't update that doc file I think. If you see the recent commits, all have tests. And the statement in that doc is not true though.

mgreenbaum:
At least I tried...

Can we pair on this to write the test?

mgreenbaum:
@thangdinh Billy said you'd be the best to review this now

tests/unit/popover_test.coffee
```diff
@@ -2,17 +2,38 @@ popover = null;
 
 module "Popover unit tests",
   setup: ->
-    Ember.run ->
-      popover = Ember.Widgets.PopoverComponent.popup
-        rootElement: '#ember-testing'
-  ,
+    return
```
thangdinh: if there is no setup, just remove the setup function

src/popover.coffee
```diff
@@ -234,8 +234,15 @@ Ember.Widgets.PopoverComponent.reopenClass
...
 
-  popup: (options) ->
-    @hideAll()
+  ###*
+   * Shows a popup with options used for PopoverComponent.
+   * @param {Object} options All options used for PopoverComponent.
+   * @param {boolean} [hideOthers=true] Indicates if other popover should be hidden when a new one
+   *    is shown. By default, it's set to true.
+  ###
+  popup: (options, hideOthers = true) ->
```
thangdinh: as we already have options in the popup function, hideOthers should be part of it instead of being a separate param

src/popover.coffee
```diff
@@ -234,8 +234,15 @@ Ember.Widgets.PopoverComponent.reopenClass
...
+    if hideOthers
+      @hideAll()
     rootElement = options.rootElement or @rootElement
```
thangdinh: I think this line will cause app crash if no option is passed in, i.e. option is null

tests/unit/popover_test.coffee
```diff
@@ -2,17 +2,38 @@ popover = null;
...
+      rootElement: '#ember-testing'
+
   sinon.spy(popover, "hide")
   wait().then ->
     keyEvent('.popover-content', 'keyup', 27)
   wait().then ->
     ok(popover.hide.calledOnce, "`hide` gets called when hitting escape")
+    popover.hide.restore()
+
+test 'Popover does not hide other popover if hideOthers is false', ->
```
thangdinh: the test name is not covering the 2 test cases, we can separate it into 2 tests here or make the name more general.

src/popover.coffee
thangdinh: And hideOthers should be part of the attribute of popover component, similar to the rootElement. And you can have the default value there. But you may want to do something like this:
```
  popup: (options) ->
    if Ember.isNone(options)
      hideOthers = @hideOthers
      rootElement = @rootElement
    else
      hideOthers = options.hideOthers or @hideOthers
      rootElement = options.rootElement or @rootElement

    if hideOthers
      @hideAll()
    popover = this.create options
    if popover.get('targetObject.container')
      popover.set 'container', popover.get('targetObject.container')
    popover.appendTo rootElement
    popover
```

mgreenbaum: We actually used to pass hideOthers as an option, and had hideOthers as an attribute of the component. Billy thought this was misleading since hideOthers is only used when launching the popup, with the value of hideOthers in the component being ignored. We pass it as a parameter of `popup

tests/unit/popover_test.coffee
mgreenbaum: `Popover respects hideOthers option`

src/popover.coffee
thangdinh: I can see what Billy came from. It also makes sense. I am not sure which way is better though. 
About the `options.hideOthers` and `@hideOthers`, when we pass the `hideOthers` into `options` in the `popup` function, it should then set the `@hideOthers` to that value.

tests/unit/popover_test.coffee
```diff
@@ -2,17 +2,38 @@ popover = null;
...
+    popover = Ember.Widgets.PopoverComponent.popup({rootElement: '#ember-testing'})
+  andThen ->
+    ok(jQueryHide.calledOnce, "Hide action is triggered")
+    jQueryHide.reset()
+
+  # Case 2: other popover should not be hidden when hideOthers is false
+  Ember.run ->
+    popover = Ember.Widgets.PopoverComponent.popup({rootElement: '#ember-testing'}, false)
+  andThen ->
+    ok(jQueryHide.notCalled, "Hide action is not triggered")
```
thangdinh: please change the message to indicate that hide action should not be triggered when hideOptions is set to false.

tests/unit/popover_test.coffee
```diff
@@ -2,17 +2,38 @@ popover = null;
...
+
+test 'Popover does not hide other popover if hideOthers is false', ->
+  expect 2
+
+  jQueryHide = sinon.stub($.fn, 'trigger');
+  # Case 1: other popover should be hidden by default
+  Ember.run ->
+    popover = Ember.Widgets.PopoverComponent.popup({rootElement: '#ember-testing'})
+  andThen ->
+    ok(jQueryHide.calledOnce, "Hide action is triggered")
```
thangdinh: please change the message to indicate that hide action should be triggered by default when hideOptions is not set.

src/popover.coffee
mgreenbaum: I'm going to keep `hideOthers` as an argument to `popover` since it's a more minimal change, and has less potential for error

src/popover.coffee
mgreenbaum: This is unrelated to my change. I don't think having default values if options is missing is a no-brainer - there are advantages to having options effectively required (enforced at runtime, although it could be improved with an explicit error message). This change should be considered independently from this PR

mgreenbaum:
I've addressed the existing code review comments @billy-addepar @thangdinh

mgreenbaum:
Not sure why the tests would fail here, I only changed some strings in the test (and removed some random build files that seemed unintentionally added) @billy-addepar @thangdinh

billy-addepar:
retest

billy-addepar:
@mgreenbaum you should run `grunt dist` after changing anything to update generated JS files. Otherwise, test will fail. I fixed the tests for you.

thangdinh:
LGTM

PR #239, opened 2016-07-01
Allow popover links to open on right clicks

embooglement:
This commit adds two new parameters to the `PopoverLinkComponent`, `openOnLeftClick` and `openOnRightClick`. Setting either of these to true will open the popover on the corresponding click event. The default behavior is to only open on left-clicks.


embooglement:
@Addepar/fire @Addepar/pants

src/popover_link.coffee
```diff
@@ -44,4 +46,18 @@ Ember.Widgets.PopoverLinkComponent = Ember.Component.extend
...
       popover.appendTo @get('rootElement')
 
+  click: ->
+    if not @get('openOnLeftClick')
+      return false
+
+    @_openPopover()
+    return false
+
+  contextMenu: ->
```
angelagu: This should be renamed to `openContextMenu` instead

embooglement: `contextMenu` is the name of the action, this can't be changed.

embooglement: Similar to `click`

angelagu:
Why would we want to disable left or right click?

angelagu:
lgtm

josheisner:
Is this what I think it is??? :o

src/popover_link.coffee
cyril-sf: http://emberjs.com/api/classes/Ember.View.html#toc_event-names

PR #241, opened 2016-07-11
make ensure-visible settable in select component

PhyFrancis:
This makes it possible to not ensure visible in selection component.

code review: @Addepar/onfense , @Addepar/pants  @thangdinh , @Addepar/fire


src/select.coffee
```diff
@@ -400,6 +400,9 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
     )
   .property 'groupedContent.[]'
 
+  # whether or not to ensure that highlighted option is visible
```
addepar-andy: i don't understand this comment

embooglement: I don't either.

embooglement:
Make sure this has a unit test

src/select.coffee
PhyFrancis: how about `If yes, the highlighted selection option is always visible. This may result in a 'jump': when there are duplicated options and mouse hovers over the second occurrence of the option, it will jump to the first occurrence.`

addepar-andy: `whether selecting an item should scroll the view to highlight the item`

addepar-andy:
actually now that i understand what's going on, i have some reservations about this.  i'll comment in the other PR

addepar-andy:
okay, this seems like a legitimate use case now

src/select.coffee
embooglement: I think we should rename this option to something more aligned with that description. Maybe something like `scrollSelectionIntoView`?

PhyFrancis: This variable controls whether the function `@ensureVisiable` is called or not, while the function `@ensureVisible` controls how to ensure visibility, it can be scrolling or other way. So the name is aligned IMO.

addepar-andy: i think "ensureVisible" is not really that clear to begin with.  naming the thing that controls when that is triggered is a step away and even less clear.

embooglement: What? Rename the function too. This name is super weird. Even just reading the description for the PR, I had no idea what this change was doing.

PhyFrancis:
test added and has passed

PhyFrancis:
bump. test has passed.

Gruntfile.coffee
```diff
@@ -33,7 +33,7 @@ module.exports = (grunt) ->
         singleRun: true
       unit:
         configFile: 'karma.conf.js'
-        singleRun: false
+        singleRun: true
```
thangdinh: We don't want this into the master. I think. @cyril-sf

src/select.coffee
```diff
@@ -409,9 +415,9 @@ Ember.AddeparMixins.ResizeHandlerMixin, Ember.Widgets.KeyboardHelper,
       value = content.objectAt index
     else # setter
       index = content.indexOf value
-      @setHighlightedIndex index, yes
+      @setHighlightedIndex index, @get('shouldEnsureVisible')
```
thangdinh: So is this a "hack" that when there are duplicated options, none of them is highlighted? I feel like we should have different way to change the way we set highlight, on an object instead of a value, which requires a different requirement for data passed in the component.

thangdinh:
as this affects the `highlighted` computed property, could you add a test to test that new behavior there?

src/select.coffee
PhyFrancis: no both of them are highlighted, the 'shouldEnsureVisible' is a flag to control whether the option with the `index` is visible in the dropdown. See the doc for the 'shouldEnsureVisible' variable.

PhyFrancis:
> as this affects the highlighted computed property, could you add a test to test that new behavior there?

Yeah the new behavior is in the test.

src/select.coffee
thangdinh: Oh got it. Thanks

thangdinh: Btw, when there is only one item, why shouldn't it jump to that highlighted item? After this change, it won't, right?

tests/unit/select_test.coffee
```diff
@@ -344,3 +344,14 @@ test 'optionValuePath with nested valuePath', ->
...
     'The right selection is retrieved'
+
+test 'shouldEnsureVisible controls whether to ensure visibility', ->
+  expect 0
+
+  select = @subject
+    content: ['foo']
+    shouldEnsureVisible: no
+    ensureVisible: ->
+      ok no, 'ensureVisible is not called if shouldEnsureVisible is false'
```
thangdinh: should we use sinon.spy instead of this, and check if the spy is not called? Having `expect 0` is not natural

src/select.coffee
PhyFrancis: no it won't. This PR is backward compat. The new behavior is only when user override the `shouldEnsureVisible` to false.

thangdinh: my bad

PhyFrancis:
bump, comments addressed, tests passed

tests/unit/select_test.coffee
```diff
@@ -344,3 +344,16 @@ test 'optionValuePath with nested valuePath', ->
...
+  expect 1
+
+  select = @subject
+    content: ['foo']
+    shouldEnsureVisible: no
+
+  spy = sinon.spy select, 'ensureVisible'
+  select.set 'highlighted', 'foo'
+
+  andThen ->
```
thangdinh: you don't need `andThen` here

tests/unit/select_test.coffee
```diff
@@ -344,3 +344,16 @@ test 'optionValuePath with nested valuePath', ->
 
   equal obj2, select.get('selection'),
     'The right selection is retrieved'
+
+test 'shouldEnsureVisible controls whether to ensure visibility', ->
+  expect 1
+
+  select = @subject
+    content: ['foo']
+    shouldEnsureVisible: no
```
thangdinh: you may want to test the default case first, where `shouldEnsureVisible` true, check the spy there. And after that, you set this to No, set highlighted, and check the spy again.

tests/unit/select_test.coffee
```diff
@@ -344,3 +344,16 @@ test 'optionValuePath with nested valuePath', ->
...
+
+  spy = sinon.spy select, 'ensureVisible'
```
thangdinh: It is also a good practice to restore the spy in teardown:
`spy.restore()`

PhyFrancis:
comments addressed

tests/unit/select_test.coffee
```diff
@@ -344,3 +344,22 @@ test 'optionValuePath with nested valuePath', ->
...
+test 'shouldEnsureVisible controls whether to ensure visibility', ->
+  expect 2
+
+  select = @subject
+    content: ['foo']
+    ensureVisible: ->
+      ok yes, 'ensureVisible is called if shouldEnsureVisible is true'
+  select.set 'highlighted', 'foo'
+
+  select = @subject
```
thangdinh: I think here you can just do `select.set('shouldEnsureVisible', true)` instead of create a new instance of it, right?

tests/unit/select_test.coffee
```diff
@@ -344,3 +344,22 @@ test 'optionValuePath with nested valuePath', ->
...
   equal obj2, select.get('selection'),
     'The right selection is retrieved'
+
+test 'shouldEnsureVisible controls whether to ensure visibility', ->
+  expect 2
+
+  select = @subject
+    content: ['foo']
+    ensureVisible: ->
+      ok yes, 'ensureVisible is called if shouldEnsureVisible is true'
```
thangdinh: Why don't you use sinon.spy for this one also instead of overriding the function? You just need to restore or reset it after the first run.

tests/unit/select_test.coffee
```diff
@@ -344,3 +344,22 @@ test 'optionValuePath with nested valuePath', ->
...
+
+  select = @subject
+    content: ['bar']
+    shouldEnsureVisible: no
+
+  spy = sinon.spy select, 'ensureVisible'
+  select.set 'highlighted', 'bar'
+  equal spy.callCount, 0,
+    'ensureVisible is not called if shouldEnsureVisible is false'
+  spy.restore()
```
thangdinh: It should be in teardown because the test may be failed in between.

PhyFrancis: how do I do that? spy is only defined in this test

thangdinh: add `spy = null` outside of the test at top of the file (like the `colorPicker` here: https://github.com/Addepar/ember-widgets/blob/master/tests/unit/color_picker_test.coffee)

Then in `teardown` check if spy is not null restore it.

tests/unit/select_test.coffee
PhyFrancis: I have to stub the content of `ensureVisible` otherwise the test will throw exception. This is the most easy-to-understand way IMO

tests/unit/select_test.coffee
PhyFrancis: not sure this is an improvement. First of all, it's not clear that each test only needs at most one spy; second of all this is a 350+ line file, while the spy is only for a 10 line test, which doesn't seem to worth a global variable.

tests/unit/select_test.coffee
thangdinh: Ah, you need to render to have the origin ensureVisible working. It's fine now.

tests/unit/select_test.coffee
thangdinh: I pasted the link is just to show the global variable usage. What I said above is a good practice is to restore it in teardown. It's not necessary that each test needs at most one spy, but you can always reuse it, if not you can have multiple of them. However, it is fine in this case because the spy is on the `select` component which will be destroy between tests.

thangdinh:
LGTM. Remember to submit the dist files

PR #242, opened 2016-08-02
Bubble Popover Link Click Events

embooglement:
A previous commit (6826e00) added the ability for popover links to choose whether or not to respond to left and right click events. These events were never bubbled. This commit bubbles events if they are ignored by the popover link.


embooglement:
@Addepar/fire @Addepar/pants review?

jordansmith42:
LGTM

PR #245, opened 2016-11-18
quickfix(build): Fixes asset locations

pzuraq:
@cyril-sf 

Moves the images and assets to be included in `dist/ember-widgets`, and makes sure all references points to that location instead.


cyril-sf:
@LydiaBarnes @raemadeline

raemadeline:
LGTM after tests pass

PR #246, opened 2017-01-25
Allow SelectComponent to send action on text change

billy-addepar:
Changes:

+ Introduce debounced text component which prevents input component from sending too many action on input change. 
+ Modify SelectComponent to allow it to propagate action up on text change.

@pzuraq @Addepar/fire @Addepar/pants


billy-addepar:
bump

pzuraq:
LGTM

PR #248, opened 2017-03-16
Adds the ability to sort content groups in the select dropdown

igillis:
@billylittlefield @pzuraq

### DIFF
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -53,9 +53,9 @@ export default Ember.Component.extend(
   optionGroupPath: '',
   optionDefaultPath: '',
   /**
-   * Comparator for sorting two groups of options. By default, sorts 
+   * Comparator for sorting two groups of options. By default, sorts
    * by alphabetical ordering of the name of the group. Alternatively, you could
-   * imagine your options are something like 
+   * imagine your options are something like
    * [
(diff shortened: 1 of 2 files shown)


pzuraq:
Sorting is not really a concern for the select component itself, it should probably be done by the external context. Is there a reason we're pushing it into here?

pzuraq:
So if possible I'd like to match the API for `ember-power-select`: http://www.ember-power-select.com/docs/groups

This is the select component we'd like to use in the future once we get to Ember 2.4 (the minimum version it supports. I don't think we should change any existing grouping logic, as that would be a breaking change and we should wait until we're ready to completely refactor the widgets, but for this change could we have the `select` component check to see if each option already has a list of `options` on it, and then use that instead of grouping itself? That way it can be handled entirely in the external context.

igillis:
That API does seem much nicer. I worry a bit about collisions with the change you proposed. I can imagine someone has content items which already have an `options` field. Maybe we can have an explicit flag to indicate we want the content to be grouped?

It is also a little weird to start supporting these two very different APIs. I think my approach fits much more cleanly into the current philosophy (though I agree the current approach is inferior to what power select went with).

igillis:
Spoke with @pzuraq offline -- we're going to go with this approach and save a larger re-thinking of the api for the future.

PR #249, opened 2017-04-05
Select component: Add a property which controls the template which is displays the select options

misquith:
The changes in this PR makes the select more customizable by adding a property which controls the template displayed when the dropdown is opened and we show a list of option. 

One example use-case is to display the list of options together with a description panel. The description panel updates as we hover over a specific option.

### DIFF
--- a/app/templates/select.hbs
+++ b/app/templates/select.hbs
@@ -10,23 +10,14 @@
     {{/if}}
   </a>
 </div>
-
-<div {{bind-attr class=":dropdown-menu :js-dropdown-menu dropdownMenuClass
-  isDropdownMenuPulledRight:pull-right"}}>
+<div {{bind-attr class=":dropdown-menu :js-dropdown-menu dropdownMenuClass isDropdownMenuPulledRight:pull-right"}}>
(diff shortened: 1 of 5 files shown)


misquith:
@Addepar/pants @Addepar/trex @pzuraq

app/templates/select.hbs
```diff
@@ -10,23 +10,14 @@
     {{/if}}
   </a>
 </div>
-
-<div {{bind-attr class=":dropdown-menu :js-dropdown-menu dropdownMenuClass
-  isDropdownMenuPulledRight:pull-right"}}>
+<div {{bind-attr class=":dropdown-menu :js-dropdown-menu dropdownMenuClass isDropdownMenuPulledRight:pull-right"}}>
```
misquith: The dropdownMenuClass property was not being read/substituted because of the new line

misquith:
Introducing the partial breaks existing unit tests if they require the templates to be explicitly specified

misquith:
bump

addepar-andy:
lgtm

PR #250, opened 2017-04-11
Fix issue where the select does not always handle a change in the content array properly

misquith:
This is a minor change to the dependent properties of groupedContent so that it mirrors the dependent property list of preparedContent. Specifically that dependent property depends on the `property` as well as `property.[]`

Use case:

The content attribute is set an empty array and then immediately to a non empty array. 

Bug: The select field does not show any content

### DIFF
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -274,6 +274,9 @@ export default Ember.Component.extend(
   // Mountain View
   //   Addepar
   //   Google
+  // Need to specify preparedContent as a dependent property to force this property
+  // to recompute when we switch an empty DS.PromiseArray with a new one
+  // @see preparedContent
   groupedContent: Ember.computed(function() {
     var content, groups, path, result;
     path = this.get('optionGroupPath');
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -274,8 +274,9 @@ export default Ember.Component.extend(
   // Mountain View
   //   Addepar
   //   Google
-  // Need to specify preparedContent as a dependent property to force this property
-  // to recompute when we switch an empty DS.PromiseArray with a new one
+  // For an unknown reason, we need to specify preparedContent as well as preparedContent.[] as a dependent properties
+  // to force this property to recompute when we set content to an empty array and then immediately
+  // set it to a non-empty array. This mirrors the dependent property list of preparedContent
   // @see preparedContent
   groupedContent: Ember.computed(function() {
     var content, groups, path, result;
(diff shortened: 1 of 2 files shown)


cyril-sf:
can you write a failing test?

addon/components/select-component.js
```diff
@@ -293,7 +296,7 @@ export default Ember.Component.extend(
       return result.pushObjects(groups[key]);
     });
     return result;
-  }).property('preparedContent.[]', 'optionGroupPath', 'labels.[]'),
+  }).property('preparedContent', 'preparedContent.[]', 'optionGroupPath', 'labels.[]'),
```
cyril-sf: There's probably no reason to have `'preparedContent` and `'preparedContent.[]`

misquith:
@cyril-sf Added a test which fails if preparedContent is not specified as a dependent property on groupedContent. Please take a look

misquith:
bump

misquith:
bump

addon/components/select-component.js
```diff
@@ -345,8 +349,7 @@ export default Ember.Component.extend(
       result.pushObjects(groupObj.members);
     });
     return result;
-  }).property('preparedContent.[]', 'optionGroupPath', 'labels.[]',
-    'groupSortFunction'),
+  }).property('preparedContent', 'preparedContent.[]', 'optionGroupPath', 'labels.[]', 'groupSortFunction'),
```
cyril-sf: You don't need `'preparedContent'` and `'preparedContent.[]'` The latter should be enough

tests/integration/select-component-test.js
```diff
@@ -444,3 +444,24 @@ test('Can specify a custom partial with listViewPartial', function(assert) {
...
+  expect(2);
+
+  var resultItemSelector = '.ember-select-result-item';
+  select = this.subject({
+    content: ['test content']
+  });
+  this.append();
+  var selectElement = select.$();
+  openDropdown(selectElement);
+  wait();
```
cyril-sf: this should not be needed

tests/integration/select-component-test.js
```diff
@@ -444,3 +444,24 @@ test('Can specify a custom partial with listViewPartial', function(assert) {
       'The specified listViewPartial is rendered');
   });
 });
+
+test('Displays options when content is changed to another empty array and then immediately assigned to a content array', function() {
```
cyril-sf: Not sure this describes well what the test want to capture. 

You want something like

```
Select handles a change in the content array properly
```

addon/components/select-component.js
misquith: The test in the PR fails if 'preparedContent' is not specified

PR #251, opened 2017-06-09
Pzuraq/upgrade/ember 1.11

pzuraq:


app/templates/select-item-layout.hbs
```diff
@@ -1,5 +1,11 @@
 {{#if view.content.isGroupOption}}
   {{view.content.name}}
 {{else}}
-  {{yield}}
```
billy-addepar: I have some concern here. This changes behavior of this template. People who use this with some custom content is now expect the change.

pzuraq: True, that's probably what caused that one regression we fixed. I don't remember why this was causing issues, I think it was because we were relying on both `templateName` and `layoutName` to work together and that doesn't work anymore. In other words, you _must_ use block syntax in a component for `{{yield}}` to work.

pzuraq: This also doesn't make any sense logically because the `select-item` is used in the View helper.

PR #252, opened 2017-07-26
handlebars line endings

addepar-andy:
Reviewers: @Addepar/ice


pzuraq:
LGTM, but we will probably need to do the HTMLBars fix to get it working for you

addepar-andy:
i'm just going to check out all files with LF endings on windows

PR #254, opened 2017-08-21
Select field should scroll to selection when opened

misquith:
This PR fixes an issue where the selected option is not visible when the select's dropdown is opened.

Steps:
Initialize a select field with a large number of options. The number of options should result in the dropdown appearing with a scroll bar.
Set the selection (selected option) to the last option (Or any option which must be scrolled to).
Open the dropdown
Bug: Although the selected option is highlighted, it is hidden. The user needs to scroll to the option. Due the another issue with respect the mouse over events, the dropdown does scroll to the selected option on hovering over the select field

@Addepar/trex @Addepar/pants @Addepar/ice

### DIFF
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -191,7 +191,15 @@ export default Ember.Component.extend(
     dropupMenuTop = dropdownButtonOffset.top - dropdownMenuHeight - dropdownMargin;
     this.set('isDropup', dropupMenuTop > window.scrollY && dropdownMenuBottom > window.innerHeight);
     this.set('isDropdownMenuPulledRight', dropdownButtonOffset.left + dropdownMenuWidth + dropdownMargin > window.innerWidth);
-    return this.$('.js-dropdown-menu').css('visibility', 'visible');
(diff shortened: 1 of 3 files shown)


addon/components/select-component.js
```diff
@@ -191,7 +191,13 @@ export default Ember.Component.extend(
     dropupMenuTop = dropdownButtonOffset.top - dropdownMenuHeight - dropdownMargin;
     this.set('isDropup', dropupMenuTop > window.scrollY && dropdownMenuBottom > window.innerHeight);
     this.set('isDropdownMenuPulledRight', dropdownButtonOffset.left + dropdownMenuWidth + dropdownMargin > window.innerWidth);
-    return this.$('.js-dropdown-menu').css('visibility', 'visible');
+    this.$('.js-dropdown-menu').css('visibility', 'visible');
+
+    // When the dropdown is opened or re-rendered, scroll to the highlighted option
+    const highlightedIndex = this.get('highlightedIndex');
+    if (highlightedIndex !== -1 && this.get('shouldEnsureVisible')) {
```
addepar-andy: you should include this as a property up top so its a bit more discoverable.

PR #255, opened 2017-10-09
Update label of selected item when selected options change

michelleshu:
This PR is a small update to the select component to update the selected item label when the contents of 'selectableOptions` change.

My current use case is a filter dropdown select where each select option contains an item count (e.g. Filter: Selected Items (5)) and the item count (the `displayName` field) may change. I am currently able to keep the `displayName` text of select options updated, but I would like the currently selected item's `displayName` to also update with this change.

@Addepar/ice

### DIFF
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -463,7 +463,10 @@ export default Ember.Component.extend(
     if (!Ember.isNone(selection)) {
       const selectedOption = 
         selectableOptions.findBy(valuePath, selection[valuePath]);
-      this.set(`selection.${labelPath}`, selectedOption[labelPath]);
+        
+      if (!Ember.isNone(selectedOption)) {
+        this.set(`selection.${labelPath}`, selectedOption[labelPath]);
+      }
     }
   }, 'selectableOptions.[]', 'showDropdown'),
 
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -462,10 +462,10 @@ export default Ember.Component.extend(
     // Update display name of selected option
     if (!Ember.isNone(selection)) {
       const selectedOption = 
-        selectableOptions.findBy(valuePath, selection[valuePath]);
-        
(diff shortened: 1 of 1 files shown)


addon/components/select-component.js
```diff
@@ -446,13 +446,28 @@ export default Ember.Component.extend(
...
       }
     }
+
+    // Update display name of selected option
+    if (!Ember.isNone(selection) && !Ember.isEmpty(labelPath) && !Ember.isEmpty(valuePath)) {
+      const selectedOption = 
+        selectableOptions.findBy(valuePath, Ember.get(selection, valuePath));
+
+      if (!Ember.isNone(selectedOption)) {
+        this.set(`selection.${labelPath}`, Ember.get(selectedOption, labelPath));
```
pzuraq: If I'm reading this correctly, it seems like `selection` and `selectedOption` are two different objects and we're syncing them. This worries me, because they _should_ be the same object, so either the select component is doing something really wrong, or actually do have two external objects and we're mutating the state of one which we probably shouldn't be doing.

michelleshu: Hmm yes, I initially thought they should be the same also, but it appears that `selection` is not the same object as the corresponding member of the `selectableOptions` array.

There is only one object passed in externally (from the array of `content` which maps to the `selectableOptions`). The `selection` is derived from the `value` parameter passed in, which is just the `optionValuePath` property of the selected option.

I'm not too familiar with what is going on internally here but I can take a closer look

pzuraq: Probably good to make sure we know what's going on here. This may still be the right solution for now, until we redo Select, but it feels like it could cause a lot of bugs if there are cases where the `selection` is the same as the `value`. If it's always a clone or something then I think we're fine.

michelleshu: @pzuraq Actually I think what is going on here is that I updated `selectableOptions` to a new array and the `selection` is still referencing an element in the old options array. I'll try to see if I can just instead update the labels directly in the original options array and if the changes will propagate

PR #256, opened 2017-11-20
Add placeholder property to typeahead

michelleshu:
@Addepar/ice 

Adding an option to add a placeholder to ember-typeahead component

### DIFF
--- a/addon/components/typeahead-component.js
+++ b/addon/components/typeahead-component.js
@@ -5,13 +5,18 @@ export default SelectComponent.extend({
   layoutName: 'typeahead',
   searchFieldClass: 'form-control',
   disabled: false,
+  placeholder: '',
 
-  searchView: Ember.TextField.extend({
-    class: 'ember-select-input',
-    valueBinding: 'parentView.query',
-    focusIn() {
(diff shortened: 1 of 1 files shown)


addon/components/typeahead-component.js
```diff
@@ -5,13 +5,18 @@ export default SelectComponent.extend({
...
-    class: 'ember-select-input',
-    valueBinding: 'parentView.query',
-    focusIn() {
-      this.set('parentView.showDropdown', true);
-    }
+  searchView: Ember.computed('placeholder', function() {
+    const placeholder = this.get('placeholder');
+    return Ember.TextField.extend({
+      class: 'ember-select-input',
+      placeholder: placeholder,
```
pzuraq: You could probably do `placeholderBinding: 'parentView.placeholder'` mirroring the way we do `valueBinding` below here, I think that would be much cleaner

michelleshu: Sounds good, I'll give that a shot!

michelleshu:
@pzuraq Does this look ok to you now?

PR #258, opened 2017-12-01
Bump `ember-list-view` version

twokul:


mixonic:
This addon no longer uses list-view as of https://github.com/Addepar/ember-widgets/pull/265

PR #262, opened 2018-03-30
Don't open dropdown on keyDown if select is disabled

michelleshu:
Fixes an issue where a disabled select would still open its dropdown on keyDown. Expected behavior is that dropdown will not open when select is disabled.

@Addepar/ice 
cc @Addepar/twex


cyril-sf:
Can you write an integration test to get that behavior covered?

michelleshu:
Sure, I'll add one

michelleshu:
Added a test, confirmed that it fails before this change.

cyril-sf:
Thanks!

PR #263, opened 2018-04-09
Select Component: Send valueChanged when query is cleared

michelleshu:
Currently the query text in select component is automatically cleared when the dropdown is hidden, via Enter key, Escape key or focus out. Select component has an existing `valueChanged` action, but it's currently only sent when a user manually types into the query text input and not when the input is cleared programmatically. 

I'd like to send the `valueChanged` action on clear, so that parent contexts can be notified that the query has changed. This is especially relevant for selects that use server side search, so they can then reissue the search with an empty query string for the next time the dropdown is opened.

@Addepar/ice 
@Addepar/twex

### DIFF
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -262,7 +262,8 @@ export default Ember.Component.extend(
         });
       // clear the query string when dropdown is hidden
       } else {
-        return this.set('value', '');
+        this.set('value', '');
+        this.get('parentView').send('valueChanged', '');
       }
     }, 'parentView.showDropdown'),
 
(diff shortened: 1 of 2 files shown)


addon/components/select-component.js
```diff
@@ -262,7 +262,8 @@ export default Ember.Component.extend(
         });
       // clear the query string when dropdown is hidden
       } else {
-        return this.set('value', '');
+        this.set('value', '');
+        this.get('parentView').send('valueChanged', '');
```
pzuraq: Shouldn't this be `this.sendAction('valueChanged', '')`?

michelleshu: Since this is within the DebouncedTextComponent child view, I'm using `this.get('parentView').send()`, to trigger the action from the SelectComponent parent view. I'm not sure this is the right way to do things, but it's currently done this way on 275-276 in `propagateNewText`

PR #265, opened 2018-05-11
Use vertical-collection

mixonic:
For ember-widgets to be usable in Ember 1.13, it must not be dependent upon list-view. This PR removes uses of list-view and replaces them with [vertical-collection](https://github.com/html-next/vertical-collection).

TODO:

* [x] Get basic vertical collection implementations to have equivalent styles
* [x] Test/implement multi select
* [x] Remove styles related to list-view
* [x] Remove list-view
* [x] Ensure the dropdown scrolls to the correct item when opened (seems to work, but there is a failing test)
* [ ] ~Add 1.12 to the build matrix, ensure it passes.~ I'm going to skip this, it seems like the test suite is fubar in 1.12. https://github.com/Addepar/ember-widgets/pull/266
* [x] Add 1.13 to the build matrix, ensure it passes.

### DIFF
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -220,8 +220,8 @@ export default Ember.Component.extend(
   selectedItemView: Ember.computed(function() {
     return this.get('itemView').extend({
       tagName: 'span',
-      labelPath: Ember.computed.alias('controller.optionLabelPath'),
-      context: Ember.computed.alias('controller.selection'),
+      labelPath: Ember.computed.alias('selectComponent.optionLabelPath'),
+      context: Ember.computed.alias('selectComponent.selection'),
       /**
       * Note: This view is an extension of the view used to display
       * each option in the dropdown list.
@@ -245,15 +245,15 @@ export default Ember.Component.extend(
   }, 'selection', 'optionLabelPath')),
 
   searchView: DebouncedTextComponent.extend({
-    placeholder: Ember.computed.alias('parentView.placeholder'),
-    valueBinding: 'parentView.query',
+    placeholder: Ember.computed.alias('selectComponent.placeholder'),
+    valueBinding: 'selectComponent.query',
     // we want to focus on search input when dropdown is opened. We need to put
     // this in a run loop to wait for the event that triggers the showDropdown
     // to finishes before trying to focus the input. Otherwise, focus when be
     // "stolen" from us.
     showDropdownDidChange: Ember.observer(function() {
       // when closing, don't need to focus the now-hidden search box
-      if (this.get('parentView.showDropdown')) {
+      if (this.get('selectComponent.showDropdown')) {
         return Ember.run.schedule('afterRender', this, function() {
           if ((this.get('_state') || this.get('state')) === 'inDOM') {
             return this.$().focus();
@@ -262,9 +262,9 @@ export default Ember.Component.extend(
       // clear the query string when dropdown is hidden
       } else {
         this.set('value', '');
-        this.get('parentView').send('valueChanged', '');
+        this.get('selectComponent').send('valueChanged', '');
       }
-    }, 'parentView.showDropdown'),
+    }, 'selectComponent.showDropdown'),
 
     /**
       Delegates to parent view (The select component) to propagate this data up.
@@ -272,7 +272,7 @@ export default Ember.Component.extend(
       @override
     */
     propagateNewText: function(newText) {
-      this.get('parentView').send('valueChanged', newText);
+      this.get('selectComponent').send('valueChanged', newText);
     },
   }),
 
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -1,5 +1,4 @@
 import Ember from 'ember';
-import ListView from 'ember-list-view';
 
 import BodyEventListener from '../mixins/body-event-listener';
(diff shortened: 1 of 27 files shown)


addon/components/select-component.js
```diff
@@ -145,7 +144,7 @@ export default Ember.Component.extend(
     } else {
       return this.get('originalItemViewClass');
     }
-  }).property('showTooltip'),
```
pzuraq: So, we definitely disabled prototype extensions in Iverson. I could have sworn we updated Ember Widgets as well, unsure how this is still here

addon/styles/select.less
```diff
@@ -359,8 +348,7 @@
 
   .dropdown-menu {
     // This causes stress
-    min-width: 200px;
-    max-width: 600px;
+    width: 200px;
```
cyril-sf: That seems weird.

@LydiaBarnes thoughts?

app/templates/multi-select.hbs
```diff
@@ -11,14 +17,25 @@
...
-  {{view listView
-    tagName="ul"
-    classNames="ember-select-results"
-    heightBinding="dropdownHeight"
-    rowHeightBinding="rowHeight"
-    contentBinding="groupedContent"
-    itemViewClassBinding="itemView"
-  }}
+  <ul class="ember-select-results" style={{collectionStyle}}>
+    {{#if showDropdown}}
```
pzuraq: This is a good hack, though it makes me wonder if we should have VC add a mutation observer or something for the parent to schedule updates. We've also talked about having a `registerAPI` action that sends up a manual API, could have a `scheduleUpdate` call so you can manually trigger a rerender.

addon/styles/select.less
LydiaBarnes: @mixonic Hmm what is the need for this change? Selects all over the app would be affected and theoretically not be able to grow, so words might get ellipsed too early, etc. A lot of them have widths hacked over them in Iverson but its hard to tell the full effect.

mixonic: Thanks for noting this @cyril-sf. @LydiaBarnes the issue is that when scrolling if wider content is present than was present during initial render, the box becomes wider. this can be pretty surprising.

I haven't confirmed what the existing behavior is, but I don't believe we want the box to be changing widths as the user scrolls.

* If this is already the behavior and the "width hacks" in Iverson already handle it, then I can just remove this change.
* If the existing behavior is that the initial width of the drop-down is retained, then I'll need to do some extra work to make that happen.

What do you think?

app/templates/multi-select.hbs
mixonic: This is indeed a hack, I'll add a comment. Basically it is because there is jQuery logic managing the dropdown open/close state.

Ideally you *never* render DOM the user does not see. I think the real fix here is simple: Stop using jQuery to manage the open/closed dropdown state. But I'm not inclined to make that change during this initial effort.

Thus, the hack.

addon/styles/select.less
LydiaBarnes: Ok I poked around at Iverson selects. Looks like we overwrite width to be 100%. I guess through some sorcery everything still works out. So theoretically if you change it here, it won't effect anything since it gets overwritten globally there.
I would make it 300 though just to be super safe, since we also globally add a min-width: 300px on them.
Eventually this nightmare select will go away right? 😀

app/templates/multi-select.hbs
LydiaBarnes: You could go ahead and switch it to our new dropdown :trollface:

addon/styles/select.less
mixonic: I'll confirm the behavior hasn't changed, or discover exactly how it has. I don't think a width of 100% would stomp the `max-width` here? I'll check.

mixonic: Ok, I did a visual side-by-side comparison to make sure things look the same.

I'm going to leave the width value. With list view if all items were < 200px wide when you scrolled down to them they would be truncated. In this implementation if there is no width specified here then the select will widen and narrow according to the content (within the min/max widths).

I think the better default here is to just pin to 200px wide. @LydiaBarnes is correct that often a width of 100% is set in Iverson, and override this, so the default is largely moot.

cyril-sf:
This is not ready to go. The bug I fixed today by removing `stateManager` makes me question why the `popover` template was changed that way.

app/templates/popover.hbs
```diff
@@ -3,5 +3,8 @@
   <h4 class="popover-title">{{title}}</h4>
 {{/if}}
 <div class="popover-content">
-  {{view view._contentViewClass}}
+  {{view _contentViewClass
+      content=content
+      isSliding=isSliding
```
cyril-sf: Are those parameters really needed?

mixonic: @cyril-sf I did some quick research. The aim of this change was definitely to move away from the implicit context from this template being available as the context of the `_contentViewClass`. `_contentViewClass` is often (always?) a `ContainerView`, and so the context is also inherited to the template of the `itemViewClass`.

Regardless of my original intent, I think you're correct that we shouldn't need to pass any values here for existing Iverson code to function on 1.12.

To avoid the inherited context stuff and remove use of `parentView` I expect we will need to pass `content` and maybe `popoverComponent` so we can update the content view class and item view classes.

I was in the weeds when I touched this and may be been on a wild goose chase wrt the interaction of popups and selects.

PR #266, opened 2018-05-14
Add 1.12 to tested builds

mixonic:
Spun out of https://github.com/Addepar/ember-widgets/pull/265


mixonic:
The test suite is basically broken against 1.12. I'm going to instead fold this back into #265 and only test 1.11 and 1.13.

PR #270, opened 2018-09-06
Don't create select rows for empty grouping names

embooglement:


addon/components/select-component.js
```diff
@@ -357,12 +357,14 @@ export default Ember.Component.extend(
     var sortedGroupObjs = groupObjs.sort(this.get('groupSortFunction'));
     var result = Ember.A();
     sortedGroupObjs.forEach(function(groupObj) {
-      result.pushObject(
-        Ember.Object.create({
-          isGroupOption: true,
-          name: groupObj.name
-        })
-      );
+      if (groupObj.name) {
```
twokul: should we add a test?

PR #271, opened 2018-09-10
WIP add new `openModal`/`openPopover` API

bantic:
This fixes issues with ContainerView on Ember 1.13 by changing the way modals are rendered to use the `{{component` template helper instead of ContainerView.

It adds a new API:
`this.get('popoverService').openModal(componentName, options)`

To do:
  [x] Add similar `openPopover` API
  [x] Test that `openPopover` allows choosing whether to hideAll others
  [x] Test that `openModal` always hides all others
  [ ] Add deprecation warnings for the old `ModalComponentClass.extend(ComponentClassSpec).popup` API

Feedback is welcome. I'll be cleaning this up a bit tomorrow.

PR #272, opened 2018-09-11
Refactor API to avoid containerView, use component template helper

bantic:
supersedes #271 

Fixes an issue with ContainerView and inadvertent view rerendering when using popovers and modals with Ember 1.13. Adds a new API so that the change can be gradually adopted.
On the exposed `popover` service, adds two new methods:
```
// Render the given component name, with the passed options as properties, in a modal:
popoverService.openModal(modalComponentName, options={});

// Render the given component name, with the passed options as properties, in a popover:
popoverService.openPopover(popoverComponentName, options={});
```

Adds some additional tests to ensure that passing properties (and `xBinding` bindings) to the modals|popovers works as expected.


addon/mixins/new-popover-modal-api.js
```diff
@@ -0,0 +1,30 @@
...
+    this._super(...args);
+    let opts = this.get('_propertyOptions');
+    let isNewApi = this.get('__renderPopoverNewAPI');
+    Ember.assert(`[ember-widgets] popover or modal must be initialized with _propertyOptions`, !isNewApi || !!opts);
+    this._setupProperties(opts);
+  },
+  _setupProperties(props={}) {
+    let boundPropKeys = [];
+    let bareProps = {};
+    let BINDING_RE = /Binding$/;
```
kybishop: Any reason to keep this a local variable over an actual `const` at the top of the file?

kybishop: Could we perhaps add a comment explaining "binding" props vs "bare" props? I'm not entirely clear on what these are being aliased to/why the aliases are necessary.

addon/services/popover.js
```diff
@@ -1,12 +1,99 @@
...
+
+    // Array of specifications for modals and/or popovers
     this.popovers = Ember.A([]);
   },
 
-  open(destinationElement, ComponentClass, options) {
+  openPopover(componentName, options={}, hideAll=true) {
+    assert(`[ember-widgets#openPopover] {{render-popover}} must be included in a template for popovers to open`, !!this._popoverDidRender);
+
+    let ComponentClass = this.container.lookupFactory(`component:${componentName}`);
```
kybishop: There are some views that get rendered via `.popover()`. Might need to pass in the whole container registry string, or at least an option to specify a view.

addon/services/popover.js
```diff
@@ -1,12 +1,99 @@
...
+      this._hideAllPopovers();
+    }
+    this.popovers.pushObject(popoverSpec);
+    return closePopover;
+  },
+
+  openModal(componentName, options={}) {
+    assert(`[ember-widgets#openModal] {{render-popover}} must be included in a template for popovers to open`, !!this._popoverDidRender);
+
+    let ComponentClass = this.container.lookupFactory(`component:${componentName}`);
```
kybishop: Not sure if modals are also occasionally views, but worth checking.

addon/components/modal-component.js
```diff
@@ -250,6 +252,8 @@ ModalComponent.reopenClass({
   hideAll() {
     return $(document).trigger('modal:hide');
   },
+
+  // TODO Deprecate this old API
```
mixonic: As I believe Addepar are the only ones using this library (there are no npm releases for example) I think we can be liberal changing the API without a deprecation phase.

addon/mixins/new-popover-modal-api.js
```diff
@@ -0,0 +1,30 @@
+export default Ember.Mixin.create({
+  init(...args) {
+    this._super(...args);
+    let opts = this.get('_propertyOptions');
+    let isNewApi = this.get('__renderPopoverNewAPI');
+    Ember.assert(`[ember-widgets] popover or modal must be initialized with _propertyOptions`, !isNewApi || !!opts);
```
mixonic: Is there a higher level error that can be thrown here? "A component extending from ember-widget's model or popover base classes cannot be invoked directly. Use the popover service."

If `this.get('__renderPopoverNewAPI')` it so be stripped it need so be part of the `assert` line:

```js
Ember.assert(
  `Some error`,
  !(this.get('__renderPopoverNewAPI')) || !!opts
);
```

addon/services/popover.js
kybishop: It's very possible I got fooled by [some aliasing](https://github.com/Addepar/Iverson/blob/master/iverson/app/views/report-widget-popover-content.js#L5). I'm digging through now to sanity check if there are any view-based popovers.

kybishop: Very happy to say that's the only one  🎉

addon/services/popover.js
kybishop: Checked the app myself, we're all good 😄

addon/components/modal-component.js
bantic: Sounds good. I'm replacing it with a `// TODO remove old API`

addon/mixins/new-popover-modal-api.js
```diff
@@ -0,0 +1,30 @@
...
+        boundPropKeys.push(key);
+      } else {
+        bareProps[key] = props[key];
+      }
+    });
+    boundPropKeys.forEach(boundKey => {
+      let barePropName = boundKey.replace(BINDING_RE, '');
+
+      // "install" binding
+      this[barePropName] = Ember.computed.alias(props[boundKey]);
```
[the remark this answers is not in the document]
bantic: So far they do seem to be fine with this change. But I'll keep this in mind if any errors do "popup" 😆

addon/mixins/new-popover-modal-api.js
bantic: Good call...made the change

addon/services/popover.js
bantic: thanks for checking!

addon/services/popover.js
bantic: @kybishop Thanks for finding that. I'm actually unclear on what that is doing. Tests seem to pass even though I haven't modified it. Do you have an idea for what to do about it?

@mixonic I pulled in the lookups for `ComponentClass` into the body of their respective assertions (in both `openModal` and `openPopover`), so those will now be stripped.

bantic: Actually, I spoke to soon about moving `lookupFactory` — we need to look at the `ComponentClass` to figure out the modal|popover destination element. So I will pull back that reference. I previously looked in the app and do not see any/many places where there is `rootElement` on the `ComponentClass`, so we can probably remove that altogether soon, after taking another pass. But I'll leave it as is for now to land this PR.

bantic:
Addressed PR feedback. @kybishop does anything need to be done as part of this PR to address your point above about the popoverView? I would prefer to merge this as-is and then circle back if/when that causes any issues in the app. If it's problematic it should be fairly obvious when refactoring to use `openPopover`.

PR #276, opened 2018-10-30
Test highlights. Clean up highlights logic.

mixonic:
~In usage with 1.13 the observer which sets the first selectable item as highlighted does not always run after the selectableOptions value has been updated. This addresses the timing issue by ensuring the observer work occurs after bindings flush.~

~Frustratingly this test does not seem to reproduce the issue (though it is nice to add anyway). In downstream usage we have another test that is resolved by the observer timing change.~

After looking further at how this fix integrates with the downstream consumer I think it is too invasive. That said, the test and code cleanup here are valid and tested against the downstream consumer, so IMO this is good to go.

### DIFF
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -444,12 +444,18 @@ export default Ember.Component.extend(
     }
     return this.set('selection', content.findProperty(defaultPath));
   }, 'content.[]'),
+
   selectableOptionsDidChange: Ember.observer(function() {
-    var highlighted;
+    /*
+     * If the dropdown is visible and the selected option is no longer
+     * contained in the list of selectableOptions, then force the first
+     * selectableOption to be highlighted.
+     */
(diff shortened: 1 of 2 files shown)


mixonic:
Updated the description here: this is just a test and cleanup now.

addon/components/select-component.js
```diff
@@ -482,17 +488,21 @@ export default Ember.Component.extend(
...
+      return content.objectAt(this.get('highlightedIndex'));
+    }
+    
+    // Setter
+    let index = value ? content.indexOf(value) : -1;
+    this.setHighlightedIndex(
+      index,
+      this.get('shouldEnsureVisible')
+    );
+    return value || [];
```
bantic: I see that the `|| []` isn't a change introduced by this PR, but is there any rationale for why `highlighted` should be an empty array?

PR #283, opened 2019-03-05
Fix a race condition between value/content

mixonic:
Previously if a value argument was set before content had been set (which may happen during initialization) then a selection value would be set matching the value. This selection would probably be wrong, especially if `optionValuePath` is also present, meaning selection is expected to be an object.

This refactors some of the current implementation to make clearer what the code intent is, and makes these changes:

* Rename setDefaultSelection to updateSelectionFromContent, and include logic that skips the work during initialization and propagated the value -> selection if needed.
* Move updateSelectionFromContent into `init` instead of `didInsertElement`. This work just needs to happen after arguments, not after rendering itself.
* Add tests that show value -> selection and selection -> value being honored when the content property changes.

### DIFF
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -52,6 +52,17 @@ export default Ember.Component.extend(
   optionValuePath: '',
   optionGroupPath: '',
   optionDefaultPath: '',
+
+  _hasInitialized: false,
+
+
+  init() {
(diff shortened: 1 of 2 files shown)


mixonic:
I believe this work is on the right track, but I've given up on trying to land it into the downstream app. Closing this for now.

PR #284, opened 2019-03-06
Close popups in capture phase

mixonic:
Popups were handling click events after the logic for the page had run. This meant sometimes you clicked on part of the page, Ember might re-render that part of the page and remove the click target, and *then* the closing logic here would run.

The closing logic confirms that the click target is still on the page, but as mentioned above the click target may not be.

Options here would be to remove the check for the target being on the page, or to handle closing the popovers on capture phase before other logic related to the click happens. I think that is the most appropriate solution, and so implemented it here.

### DIFF
--- a/addon/mixins/body-event-listener.js
+++ b/addon/mixins/body-event-listener.js
@@ -41,11 +41,11 @@ export default Ember.Mixin.create({
       });
     };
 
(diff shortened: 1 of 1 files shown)


mixonic:
closed in favor of https://github.com/Addepar/ember-widgets/pull/285

PR #286, opened 2019-03-06
WIP: Close popups in without checking target attachment

mixonic:
Popups are handling click events after the logic for the page has rendered. This means sometimes you click on part of the page, Ember might re-render that part of the page and remove the click target, and *then* the closing logic in body-event-listener will run.

The closing logic confirmed that the click target is still on the page, but as mentioned above the click target may not be. It isn't clear why this logic exists in the first place.

Options to rationalize this: Remove the check for the target being on the page, or handle closing the popovers on capture phase before other logic related to the click happens, or use a psuedo-before-click event like mousedown.

Here the check is simply removed, meaning clicking on something outside the box will always cause a popover to be destroyed regardless of the target's attachment state.

References:

* https://github.com/Addepar/ember-widgets/pull/285
* https://github.com/Addepar/ember-widgets/pull/284

### DIFF
--- a/addon/mixins/body-event-listener.js
+++ b/addon/mixins/body-event-listener.js
@@ -24,24 +24,19 @@ export default Ember.Mixin.create({
     }
 
     this._clickHandler = function(event) {
-      return Ember.run(function() {
+      Ember.run(function() {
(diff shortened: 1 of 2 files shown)


mixonic:
I think this is also on the right track, but I've given up on trying to land it into the downstream app. Closing this for now.

PR #288, opened 2019-03-11
Add collapsible grouped headers to select-component

jiayingxu:
Add ability to collapse/expand group headers in `select-component` and `multi-select-component`. `optionGroupPath` should be set and `isGroupHeaderCollapsible=true` on the select.

There were 2 possible approaches to implementing the hiding of the list items after the header is collapsed
1. Use CSS `display: none` to hide the list item if the header is collapsed
2. Remove the list item from the backing JavaScript data structure for the select list

Approach 1 was chosen was approach 2 was not feasible. `vertical-collection` is used to render the list items to utilize its occlusion performance capability. There currently is not a built-in mechanism in `vertical-collection` to schedule a re-measure or a re-render.

collapsible-group-headers

### DIFF
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -381,7 +381,7 @@ export default Ember.Component.extend(
       }
     });
     return result;
-  }).property('preparedContent', 'preparedContent.[]', 'optionGroupPath', 'labels.[]', 'groupSortFunction', 'collapsedGroupHeaders.[]'),
+  }).property('preparedContent', 'preparedContent.[]', 'optionGroupPath', 'labels.[]', 'groupSortFunction', 'isGroupHeaderCollapsible', 'collapsedGroupHeaders.[]'),
 
   // Array that contains the names of all grouped headers that are collapsed
   collapsedGroupHeaders: Ember.A(),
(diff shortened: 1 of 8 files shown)


addon/components/select-component.js
```diff
@@ -356,19 +360,31 @@ export default Ember.Component.extend(
...
         );
       }
-      result.pushObjects(groupObj.members);
+      if (!isGroupHeaderCollapsed) {
+        result.pushObjects(groupObj.members);
+      }
     });
     return result;
-  }).property('preparedContent', 'preparedContent.[]', 'optionGroupPath', 'labels.[]', 'groupSortFunction'),
+  }).property('preparedContent', 'preparedContent.[]', 'optionGroupPath', 'labels.[]', 'groupSortFunction', 'collapsedGroupHeaders.[]'),
```
kpfefferle: Should this also observe `isGroupHeaderCollapsible`? And while we're at it, should we move these dependent keys to the beginning of the computed definition in line with newer code?

rondale-sc: I was thinking the same thing.  The prototype extensions feel quite dated, but perhaps this is just to follow along with existing style.

rondale-sc:
@jiayingxu LGTM - does the occlussion bit mean that expanded option groups are occluded at a certain point during the scroll?  Like if I expande a section then say 2 more would the initial one that could be very far down the scroll's state revert to collapsed?

jiayingxu:
@rondale-sc, yep, the collapsing/expansing of the group headers work with occlusion. As the user scrolls up and down in the select list changing the occlusion viewport, the group headers that are collapsed or expanded are retained.

PR #290, opened 2019-03-25
Add `customFilter` option to `select-component` to supply an additional filter on each list item

jiayingxu:
Add overridable `customFilter` function to `select-component` to allow an additional filter on each list item to be applied. The default implementation of the `customFilter` function is to filter out all items that are `!isNone(item)`.

This will enable the extension of `select-component` to add an additional custom filter to the select list. The `select-component` already has built-in search capability that narrows the list of content down based on the query entered in the textbox . Building on the [pre-existing implementation](https://github.com/Addepar/ember-widgets/blob/7515e048176958c81341389ecb58797e1d553ff9/addon/components/select-component.js#L319), a second filter for `filteredContent` was added. It was discussed whether or not to add functionality to allow a list of filters to be passed into `select-component`, but ultimately we wanted the implementation to preserve the pre-built in capability of the `matcher` function.

### DIFF
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -307,17 +307,15 @@ export default Ember.Component.extend(
     } else {
       observableString = 'content.[]';
     }
+    let listItemFilter = this.get('_listItemFilter');
     ContentProxy = Ember.ObjectProxy.extend({
       _select: null,
       content: Ember.computed.alias('_select.content'),
       query: Ember.computed.alias('_select.query'),
       filteredContent: Ember.computed(function() {
-        var query, selectComponent;
-        selectComponent = this.get('_select');
-        query = this.get('query');
-        return (this.get('content') || []).filter(function(item) {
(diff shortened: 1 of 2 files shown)


addon/components/select-component.js
```diff
@@ -307,17 +307,14 @@ export default Ember.Component.extend(
...
       query: Ember.computed.alias('_select.query'),
       filteredContent: Ember.computed(function() {
-        var query, selectComponent;
-        selectComponent = this.get('_select');
-        query = this.get('query');
-        return (this.get('content') || []).filter(function(item) {
-          return selectComponent.matcher(query, item);
-        });
+        var query = this.get('query');
+        return (this.get('content') || []).filter(item => listItemFilter(query, item));
```
mixonic:
```suggestion
        let content = this.get('content');
        return content ? content.filter(item => listItemFilter(query, item)) : [];
```

No need to call `filter` on an empty array.

PR #292, opened 2019-03-27
Expose popover-modal

mixonic:
Permit non-popover non-modal flows by exposing the popover-modal base class and methods on the popover service which allow opening.

TODO:

* [x] Make sure it works downstream.
* [x] Add documentation.

### DIFF
--- a/tests/integration/components/render-popover-new-api-test.js
+++ b/tests/integration/components/render-popover-new-api-test.js
@@ -18,6 +18,15 @@ function openModal(context, componentSpec, options={}) {
   return Ember.run(() => context.container.lookup('service:popover').openModal(componentName, options));
 }
 
+function openPopoverModal(context, componentName, options={}) {
+  return Ember.run(() => {
+    let popoverService = context.container.lookup('service:popover');
+    let { popoverSpec, close } = popoverService.buildPopoverSpec(componentName, document.body, options);
+    popoverService.openPopoverSpec(popoverSpec);
+    return close;
+  });
+}
+
 moduleForComponent(
   'render-popover',
   'Integration | Component | render popover (new API)',
@@ -33,4 +42,4 @@ moduleForComponent(
   }
 );
 
-runPopoverTests(test, {makeComponent, openPopover, openModal});
+runPopoverTests(test, {makeComponent, openPopover, openModal, openPopoverModal});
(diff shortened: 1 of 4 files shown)


tests/integration/components/render-popover-new-api-test.js
```diff
@@ -18,6 +18,15 @@ function openModal(context, componentSpec, options={}) {
   return Ember.run(() => context.container.lookup('service:popover').openModal(componentName, options));
 }
 
+function openPopoverModal(context, componentName, options={}) {
+  return Ember.run(() => {
+    let popoverService = context.container.lookup('service:popover');
+    let { popoverSpec, close } = popoverService.buildPopoverSpec(componentName, document.body, options);
```
rondale-sc: In this PR what does a popoverSpec mean, I see that it is an object, but what does it achieve in this context?

mixonic: I added documentation that attempts to address this questions: https://github.com/Addepar/ember-widgets/pull/292/files#diff-4509ec26f384db2ae721a20cf7990a8fR62

The name `popoverSpec` is a bit unfortunate and perhaps we can revisit it. A `popoverSpec` is an object which is eventually rendered by the `{{render-popover}}` component. It is effectively a wrapper around a component name, arguments to a component, a method to close the component, and a destination element. Basically everything you need to render something with `{{render-popover}}`

mixonic:
This is ready for some review!

tests/integration/components/render-popover-new-api-test.js
```diff
@@ -18,6 +18,15 @@ function openModal(context, componentSpec, options={}) {
   return Ember.run(() => context.container.lookup('service:popover').openModal(componentName, options));
 }
 
+function openPopoverModal(context, componentName, options={}) {
+  return Ember.run(() => {
```
kpfefferle: Might we want to take this opportunity to update to `import { run } from '@ember/runloop';` here? I see `Ember.run` is used elsewhere in this file, but maybe we should get away from the Ember global in all 3 cases?

mixonic: @kpfefferle I don't believe this codebase is using the babel transform for the Ember module API

kpfefferle: Oh dang, good call 😬

PR #293, opened 2019-04-02
Change implementation of tab view to a tab component

jiayingxu:
Move from using views to components for `tabView`. This is to aid the upgrade path to Ember 2.x


mixonic:
🙇 Thank you for adding test coverage for this change! 🎉

PR #297, opened 2019-05-23
EditableLabel uses the component syntax instead of the view one

cyril-sf:
This is likely old code from when this library was relying only on views.


twokul:
the change is straightforward but I would appreciate a good description for future and maybe less experienced code travellers 🗺

mixonic:
It seems like `layoutName` could have been dropped entirely. This is an improvement regardless.

PR #298, opened 2019-07-12
Add ability to display number of items under group header

jiayingxu:
Add ability to display number of items under group header using `numMembers` property

Example usage:
image

### DIFF
--- a/addon/components/select-component.js
+++ b/addon/components/select-component.js
@@ -372,7 +372,8 @@ export default Ember.Component.extend(
           Ember.Object.create({
             isGroupOption: true,
             isGroupHeaderCollapsed: isGroupHeaderCollapsed,
-            name: groupObj.name
+            name: groupObj.name,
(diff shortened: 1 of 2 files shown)


kpfefferle:
Design question: should the continent names here be left-aligned? It appears from the screen capture that they are centered within an area of varying width which would likely lead to them not really looking aligned at all?

jiayingxu:
I fixed the alignment in the screenshot. The screenshot is just of a demo I created locally where I hard-coded the number of items under the header with the `countries` label in `select-item-layout`. Overriding the default layout would require creating a view that overrides `select-option`.

mixonic:
I *think* these are valid questions for the use case, but I'm not sure if they apply here in particular. Just passing the number through seems fine:

* The alignment should probably also be stable regardless of collapsed/expanded? So if you click the collapse toggle the test won't jump around?
* What does this look like if the group name is long enough to overlap the count?

PR #299, opened 2019-07-24
Apply a fix for Ember 1.13 when a select component group is expanded/collapsed

mixonic:
Add a test exercising expansion and collapse of groups. This should fail on 1.13, and I'll follow up with the fix.

### DIFF
--- a/tests/integration/select-component-test.js
+++ b/tests/integration/select-component-test.js
@@ -47,7 +47,7 @@ moduleForComponent('select-component', '[Integration] Select component', {
 
   teardown: function() {
     Ember.run(function() {
-      return select.destroy();
+      select.destroy();
     });
     Ember.run(app, 'destroy');
     select = null;
@@ -65,7 +65,7 @@ test('Test continuous queries in a row', function(assert) {
   assert.equal(select.get('filteredContent')[2], 'baz');
   select.set('query', 'bar');
   assert.equal(select.get('filteredContent')[0], 'bar');
-  return assert.equal(select.get('filteredContent')[1], 'barca');
+  assert.equal(select.get('filteredContent')[1], 'barca');
 });
 
 test('Test filtered content using array proxy', function(assert) {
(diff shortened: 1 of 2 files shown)


cyril-sf:
Can we rename the PR to explain that it impacts the Select component?

mixonic:
Done!

tests/integration/select-component-test.js
```diff
@@ -672,7 +673,62 @@ test('Dropdown does not open on key event when select is disabled', function(ass
...
+  function data() {
+    return [
+      {name: 'Sparrow', sound: 'Squawk'},
+      {name: 'Crow', sound: 'Squawk'},
+      {name: 'Dog', sound: 'bark'},
+      {name: 'Wolf', sound: 'Bark'},
+      {name: 'Sea Lion', sound: 'Bark'}
+    ];
+  }
+  function renderedItems(selectElement) {
```
cyril-sf: How different is it from `getOptionsInSelect` in `tests/helpers/components/select-helpers.js`?

https://github.com/Addepar/Iverson/blob/5382b4ea68583a521838efff83fef8d8ce26cb91/iverson/tests/helpers/components/select-helpers.js#L288-L296

tests/integration/select-component-test.js
```diff
@@ -672,7 +673,62 @@ test('Dropdown does not open on key event when select is disabled', function(ass
...
+    assert.deepEqual(
+      renderedItems(selectElement),
+      ['Bark', 'Sea Lion', 'Wolf', 'Squawk', 'bark', 'Dog'],
+      'bark expanded to render Dog'
+    );
+    click('li:contains(Bark) div');
+  });
+
+  return andThen(() => {
+    assert.deepEqual(
```
cyril-sf: Is it needed to wrap the assertions in `andThen`?

tests/integration/select-component-test.js
mixonic: 😉 different codebase. But it looks like the same thing, yes.

tests/integration/select-component-test.js
mixonic: Yeah, the `click` may have resulted in `rAF` frames from vertical collection, since this is a select (which uses vertical collection). They might not be hard required in this case, but in general the this test should be async-friendly, permitting `rAF` and such if it was needed.

tests/integration/select-component-test.js
cyril-sf: lol, sorry

PR #311, opened 2019-09-23
[DEPRECATIONS] Remove deprecations raised with Ember 1.13

cyril-sf:
Add `ember-cli-deprecation-workflow` to silence known deprecations.

Also update default to Ember 1.13.

The remaining deprecations are mostly linked to `Ember.View`.

### DIFF
--- a/addon/mixins/popover.js
+++ b/addon/mixins/popover.js
@@ -42,9 +42,11 @@ export default Ember.Mixin.create(StyleBindingsMixin, BodyEventListener, {
   }).property('contentViewClass'),
   didInsertElement: function() {
     this._super();
-    this.snapToPosition();
-    this.set('visibility', 'visible');
-    return this.set('isShowing', true);
+    Ember.run.schedule('afterRender', this, () => {
+      this.snapToPosition();
+      this.set('visibility', 'visible');
+      this.set('isShowing', true);
+    });
   },
   willDestroyElement: function() {
     this.$().off($.support.transition.end);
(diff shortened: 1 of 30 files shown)


addon/mixins/popover.js
```diff
@@ -42,9 +42,11 @@ export default Ember.Mixin.create(StyleBindingsMixin, BodyEventListener, {
   }).property('contentViewClass'),
   didInsertElement: function() {
     this._super();
-    this.snapToPosition();
-    this.set('visibility', 'visible');
-    return this.set('isShowing', true);
+    Ember.run.schedule('afterRender', this, () => {
```
bantic: This fixed a double-render warning?

cyril-sf: Yes. Modifying properties in `didInsertElement`

tests/dummy/app/app.js
```diff
@@ -15,7 +15,7 @@ App = Ember.Application.extend({
 });
 
 // Legacy view-as-component weirdness. Don't ever ever ever do this.
-Ember.Handlebars.helper('carousel-item', CarouselItem);
+Ember.Helper.helper('carousel-item', CarouselItem);
```
mixonic: 🤔 this works?! `Ember.Helper.helper` returns a helper function, usually in a file in `app/helpers/` and it can then be invoked in a template.

I don't think this is doing anything. Or if it is, wow.

Downstream we always use `{{#view 'carousel-item'}}` since `carousel-item` is in the views folder. Can the invocations of `{{carousel-item}}` in the dummy test app just be converted to that usage? Then this can certainly be deleted.

cyril-sf: I need to check the dummy app/help site.

It may work because the template and view doesn't do much. I'm also not sure that this is actually tested.

cyril-sf: Fixed

PR #312, opened 2019-09-23
[CHORE] Drop support for Ember 1.11 and 1.12

cyril-sf:


.travis.yml
```diff
@@ -26,16 +26,13 @@ env:
     # - EMBER_TRY_SCENARIO=ember-release
     # - EMBER_TRY_SCENARIO=ember-beta
     # - EMBER_TRY_SCENARIO=ember-canary
-    - EMBER_TRY_SCENARIO=ember-1.11
-    - EMBER_TRY_SCENARIO=ember-1.12
     - EMBER_TRY_SCENARIO=ember-1.13
     - EMBER_TRY_SCENARIO=ember-2.0
-    - EMBER_TRY_SCENARIO=ember-default
+    # - EMBER_TRY_SCENARIO=ember-default
```
mixonic: Why this? Testing the default setup seems quite valid?

cyril-sf: I may have been overzealous.

I was trying to limit the work done by Travis that we may not care about.

mixonic: I'd like to see it come back. The default dependencies should definitely pass, or out of the box when you clone the repo the suite may be red.

```suggestion
    - EMBER_TRY_SCENARIO=ember-default
```

cyril-sf: I didn't mean to not bring it back, I was dealing with another project

PR #179, opened 2015-02-18
Remove `TextEditor`

cyril-sf:
Its behavior is flaky and the code hard to maintain.

@igillis @addepar-andy @Dylnuge

PR #315, opened 2019-10-08
[CHORE] Remove warning around style binding

cyril-sf:
`WARNING: Binding style attributes may introduce cross-site scripting vulnerabilities; please ensure that values being bound are properly escaped. For more information, including how to disable this warning, see http://emberjs.com/deprecations/v1.x/#toc_binding-style-attributes.`

Also remove the `footer-chooser-item` template which was used by the text editor introduced in c265d6c
This component has been removed in #179

### DIFF
--- a/app/templates/font-chooser-item.hbs
+++ b/app/templates/font-chooser-item.hbs
@@ -1,3 +1,0 @@
-<div style={{view.style}}>
-  {{view.label}}
-</div>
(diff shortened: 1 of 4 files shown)


app/templates/font-chooser-item.hbs
```diff
@@ -1,3 +0,0 @@
-<div style={{view.style}}>
```
mixonic: This just seems unused? I looked for `font-chooser-item` in the codebase and in Iverson and I agree it isn't used at a glance.

cyril-sf:
@mixonic I updated the title. While it's not blocking for the upgrade to Ember 2.0, it made very difficult to parse the log/console.

This PR will allow to go 382 occurrences in the log/console to just 32.

PR #316, opened 2019-10-11
[WIP]

cyril-sf:


mixonic:
Anything in afterRender can't be causing the double-render error. By definition that error requires something is synchronously changes during the rendering phase (which is in the render queue)

The other two places here that actions queue is used look like they could be a more relevant fix though, under the presumption a focus event is triggered synchronously during rendering. They seem more germane?

bantic:
@cyril-sf Can you point out (privately if need be) where are the issues in our application that this fixes?

bantic:
I spent some time trying to better understand when the focusin/focusout events are fired (in general, not in Ember specifically). It looks like it is possible for these to interleave if the DOM changes in a way that modifies focus, for instance by removing the focused element from the DOM, or calling `element.focus()` — new focusin/out events will synchronously happen. I was able to recreate the double-render issue using a select component that renders a button, where the button's action hides the dropdown. Clicking the button triggers multiple synchronous calls to focusOut and focusIn: the button gets focus itself, which triggers focusOut and then focusIn on the select-component, and then the dropdown is hidden (due to the button's action), which removes the button from DOM, so it loses focus which then triggers another synchronous focusOut as it is removed.
Scheduling these into actions seems like a good call.

I'm not clear on what moving the popover's `didInsertElement` work from `afterRender` to `actions` fixes — were you able to figure out what the underlying issue there is?

cyril-sf:
@bantic Agreed. What seems to not make possible to rely on CSS at the moment is the way we decide if the `SelectComponent` has the focus

https://github.com/Addepar/ember-widgets/blob/ebcc0e26b3a1bf92b82dfdb2c51e6432237e239b/addon/components/select-component.js#L543_L552

addon/components/select-component.js
```diff
@@ -540,14 +540,12 @@ export default Ember.Component.extend(
     }
   }),
 
-  setFocus: function() {
-    var activeElem, selectComponent;
-    activeElem = document.activeElement;
-    selectComponent = this.$()[0];
-    if (selectComponent.contains(activeElem) || selectComponent === activeElem) {
-      return this.set('hasFocus', true);
+  setFocus: function(targetElement = document.activeElement) {
```
mixonic: When does `setFocus` not have an argument?

PR #320, opened 2019-11-13
Ignore resize events that aren't on window

bantic:
We've noticed excessive "resize" events in some app usage that came when jquery-ui.resizable was in use on the page. Because these event handlers are added with jQuery, the ["resize" events](https://api.jqueryui.com/resizable/#event-resize) that resizable fires will be bubbled to the window.

This results in a lot of unnecessary work — the select-component does not need to call [`updateDropdownLayout`](https://github.com/Addepar/ember-widgets/blob/ebcc0e26b3a1bf92b82dfdb2c51e6432237e239b/addon/components/select-component.js#L165-L205) unless the window size has actually changed.


bantic:
Another option would have been to simply use `window.addEventListener('resize'`, which, because it is not attached using jQuery, won't receive the jquery-bubbled events. But this seems the more conservative change.

PR #328, opened 2020-04-28
ADPR-61646 prevent loading text and no results text from showing up s…

seth-addepar:
…imultaneously

Before: 
image

After:
image

image

### DIFF
--- a/tests/integration/components/select-component-test.js
+++ b/tests/integration/components/select-component-test.js
@@ -1,6 +1,7 @@
 import hbs from 'htmlbars-inline-precompile';
 import { moduleForComponent, test } from 'ember-qunit';
-import { click } from '@ember/test-helpers';
+import { click, fillIn } from '@ember/test-helpers';
+import Ember from 'ember';
 
 class SelectPageObject {
   constructor(elementLookup) {
@@ -15,6 +16,18 @@ class SelectPageObject {
     return this.$('.ember-select-choice')[0];
   }
 
+  get contentLoadingElement() {
+    return this.$('.ember-select-loading')[0];
+  }
+
+  get contentEmptyElement() {
+    return this.$('.ember-select-empty-content')[0];
+  }
+
+  get noResultsElement() {
+    return this.$('.ember-select-no-results')[0];
+  }
+
   selectOptionElementContaining(content) {
     return this.$(`li:contains(${content}) div`)[0];
   }
@@ -27,8 +40,8 @@ class SelectPageObject {
     return this.selectOptionElements.map(e => e.textContent.trim());
   }
 
-  inputSearchText(text) {
-    return this.$('.ember-select-search > input').text(text);
+  async inputSearchText(text) {
+    await fillIn(this.$('.ember-select-search > input')[0], text);
   }
 }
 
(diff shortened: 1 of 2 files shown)


tests/integration/components/select-component-test.js
```diff
@@ -133,3 +138,24 @@ test('It displays the specified component when componentNameForGroupTooltip is p
     'the groupItem argument is passed'
   );
 });
+
+test('It does not display loading text and empty content component at the same time', async function(assert) {
```
seth-addepar, line 142: this test probably doesn't pass, will run locally and update

seth-addepar:
@Addepar/ap-calc @Addepar/web-core

tests/integration/components/select-component-test.js
```diff
@@ -133,3 +138,24 @@ test('It displays the specified component when componentNameForGroupTooltip is p
...
+      content=content
+      emptyContentView='some-component'
+  }}`);
+
+  await this.helpers.inputSearchText('no match');
+
+  let loadingIsVisible = this.$('.ember-select-loading').length;
+  let noResultsIsVisible = this.$('.ember-select-empty-content').length;
+
+  assert.ok(loadingIsVisible ^ noResultsIsVisible, 'Either loading or no results should be visible, but not both');
```
mixonic, line 160: Oh buddy, clever, but I think as this is a deterministic state (which one is visible should always be the same) it could be two assertions instead. On that a selector is matched, one that a selector is not matched.

tests/integration/components/select-component-test.js
```diff
@@ -133,3 +138,24 @@ test('It displays the specified component when componentNameForGroupTooltip is p
     'the groupItem argument is passed'
   );
 });
+
+test('It does not display loading text and empty content component at the same time', async function(assert) {
+  this.container.register('template:components/some-component', hbs`<span data-test-some-component>No results</span>`);
```
mixonic, line 143: Yeah I'm not sure this is enough mocking. As `emptyContentView` is used with `{{view` and not with the `{{component` helper I expect you need to register a view and template.

tests/integration/components/select-component-test.js
```diff
@@ -133,3 +138,24 @@ test('It displays the specified component when componentNameForGroupTooltip is p
...
+  {{select-component
+      content=content
+      emptyContentView='some-component'
+  }}`);
+
+  await this.helpers.inputSearchText('no match');
+
+  let loadingIsVisible = this.$('.ember-select-loading').length;
+  let noResultsIsVisible = this.$('.ember-select-empty-content').length;
+
```
NicholasMohr, line 159: make sure that your test fails before your change, ideally in your test, `isLoading` and `hasNoResults` should both true and then make sure that the loading message shows up but not no results

tests/integration/components/select-component-test.js
```diff
@@ -133,3 +151,62 @@ test('It displays the specified component when componentNameForGroupTooltip is p
...
+    await click(this.helpers.selectChoiceElement);
+    await this.helpers.inputSearchText('no match');
+
+    assert.ok(!this.helpers.contentLoadingElement, 'Loading ui is not present');
+    assert.ok(!this.helpers.contentEmptyElement, 'Empty content is not present');
+    assert.ok(!!this.helpers.noResultsElement, 'No results message is present');
+  }
+);
+
+test(
```
seth-addepar, line 198: @NicholasMohr The below test fails before the change and passes afterwards. The above two tests pass either way, and are really just backfilling tests for existing functionality

seth-addepar:
@mixonic I looked through iverson for usages of the `emptyContentView` and they're all essentially the same kind of "No results found" message. See my comment on https://addepar.atlassian.net/browse/ADPR-61646 and Angel's response.

tests/integration/components/select-component-test.js
```diff
@@ -139,23 +152,61 @@ test('It displays the specified component when componentNameForGroupTooltip is p
...
+    await this.render(hbs`
+    {{select-component
+        content=content
+        emptyContentView='some-view'
+    }}`);
+
+    await click(this.helpers.selectChoiceElement);
+
+    assert.ok(!this.helpers.contentLoadingElement, 'Loading ui is not present');
+    assert.ok(!!this.helpers.contentEmptyElement, 'Empty content is present');
```
NicholasMohr, line 172: why not just use something like `assert.dom('.portal-container').exists('message')` or `assert.dom('.portal-container').doesNotExist('message')`?
ditto for all the other occurrences here where we need to turn jquery into a boolean

NicholasMohr: or if the page object above is a legit pageobject you can just use pageObject stuff, documented here http://ember-cli-page-object.js.org/docs/v1.15.x/

tests/integration/components/select-component-test.js
NicholasMohr: I would think that the select in this case doesn't have `hasNoResults` and `isLoading` true at the same time, see my overall PR comment

tests/integration/components/select-component-test.js
seth-addepar: I paired with Matt B on this so I'll let him answer this one

tests/integration/components/select-component-test.js
seth-addepar: I think you're right. In fact, the bug being fixed is actually a case where `isLoading` and `contentIsEmpty` are true at the same time. I'm happy we now have tests for both combinations but the `contentIsEmpty` one is the culprit here.

tests/integration/components/select-component-test.js
mixonic: I don't think this repo has https://github.com/simplabs/qunit-dom installed, and so `assert.dom` is not available.

This page object is just a class following the page object pattern. We didn't introduce a dependency on ember-cli-page-object.

PR #329, opened 2020-05-05
Use #each instead of a ContainerView when rendering popovers

mmun:
This PR refactors the render-popover component to use an `#each` block rather than a `ContainerView` in order to avoid a `childViews` deprecation.

### DIFF
--- a/addon/services/popover.js
+++ b/addon/services/popover.js
@@ -131,7 +131,7 @@ export default Ember.Service.extend({
     };
 
     popoverSpec.childViews = [
-      ComponentClass.create({
+      ComponentClass.extend({
         ...options,
         closePopover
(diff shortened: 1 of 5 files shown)


addon/services/popover.js
```diff
@@ -131,7 +131,7 @@ export default Ember.Service.extend({
     };
 
     popoverSpec.childViews = [
-      ComponentClass.create({
+      ComponentClass.extend({
```
mmun, line 134: This change is required because the `{{view}}` helper sets up the `container`, `ownerView`, etc for View classes not instances, unlike a ContainerView's `childViews`.

This was causing a few tests too failing when using the deprecated API.

PR #330, opened 2020-05-08
Remove less.

aschenoni:
Convert existing less files to their corresponding compiled css. 

Ripped the bandaid off, decided to leave compiled css as is to limit the chance of introducing a breaking change. If / when these components get touched again, we may want to reference the old less variables but since these components are infrequently adjusted I don't think it is significantly important to keep the variables hanging around.


mixonic:
Wait what?

PR #331, opened 2020-05-14
Use ember-object-set-return-value codemod to cleanup set return value usage

mmun:
I used an internal codemod [ember-object-set-return-value](https://github.com/Addepar/iverson-codemods/blob/master/transforms/ember-object-set-return-value/README.md) to identify usages of `set`'s return value and then cleaned these cases up manually.

I don't think any of these actually are necessary to remove for upgrading to Ember 2, but it does provide peace of mind.

### DIFF
--- a/tests/integration/select-component-render-test.js
+++ b/tests/integration/select-component-render-test.js
@@ -507,7 +507,7 @@ test('Selected option is visible when dropdown is opened', function(assert) {
   this.render();
   // The highlighted property is set when the user hovers over the select field
   // Lets programatically set it after rendering to emulate that behavior.
-  andThen(() => select.set('highlighted', selection));
+  andThen(() => { select.set('highlighted', selection); });
   var selectElement = select.$();
   openDropdown(selectElement);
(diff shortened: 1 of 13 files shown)


tests/integration/select-component-render-test.js
```diff
@@ -507,7 +507,7 @@ test('Selected option is visible when dropdown is opened', function(assert) {
   this.render();
   // The highlighted property is set when the user hovers over the select field
   // Lets programatically set it after rendering to emulate that behavior.
-  andThen(() => select.set('highlighted', selection));
+  andThen(() => { select.set('highlighted', selection); });
```
twokul, line 510: that does not seem related, does it?

mmun: The codemod picks it up because it's passed a return value of the callback.

mmun: I manually fixed this, otherwise it would have been changed to

```js
andThen(() => (select.set('highlighted', selection), select));
```

